PR #28, opened 2023-08-26
update controller runtime to 0.16.0

haouc:
**What type of PR is this?**
Controller-runtime just released new version 0.16.0 which includes a few important changes. One of them ensures health servers start before cache starts, which can avoid failure on liveness check upon startup and thus pod gets terminated in some environments.

Liveness probe may fail and container may restart if controller takes too long to start

Upgrade controller-runtime to v0.16.0 to take advantage of the latest improvements.

**Automation added to e2e**:

```release-note

```


pkg/config/runtime_config.go
```diff
@@ -82,13 +84,22 @@ func BuildRestConfig(rtCfg RuntimeConfig) (*rest.Config, error) {
 
 // BuildRuntimeOptions builds the options for the controller runtime based on config
 func BuildRuntimeOptions(rtCfg RuntimeConfig, scheme *runtime.Scheme) ctrl.Options {
+	// if DefaultNamespaces in Options is not set, cache will watch for all namespaces
```
jdn5126, line 87: I think I am misreading this. We want VPC RC client to watch all namespaces, right? Since this is used by pod webhook

haouc: We don't set namespace here unless users deploy the controller to dataplane and specifically set the flag `rtCfg.WatchNamespace` otherwise it is default to `corev1.NamespaceAll` (https://github.com/aws/amazon-network-policy-controller-k8s/blob/6d4852c37dfa8c52ad90c8d59e4207f3d6e553f3/pkg/config/runtime_config.go#L61). So in general case, it is all namespaces.

jdn5126: Ah, I was confusing which repo this was

jdn5126 (approved):

PR #44, opened 2023-10-13
update go version to 1.21

haouc:
**What type of PR is this?**
Upgrade go version to 1.21

go vulns

We need mitigate the discovered go vulns

**Automation added to e2e**:

no

no
```release-note

```


orsenthil (approved):
LGTM. Only runtime and core-libraries are updated.

PR #58, opened 2023-12-01
fix ingress dest ports issue

haouc:
**What type of PR is this?**
Fixing an issue for ingress port in policyendpoints

bug
#52 

The controller shouldn't use src port in PE's ingress endpoint.

**Automation added to e2e**:

```release-note

```


achevuru:
lgtm. Maybe, let's add an UT case for this scenario here - https://github.com/aws/amazon-network-policy-controller-k8s/blob/main/pkg/resolvers/endpoints_test.go#L185

haouc:
> lgtm. Maybe, let's add an UT case for this scenario here - https://github.com/aws/amazon-network-policy-controller-k8s/blob/main/pkg/resolvers/endpoints_test.go#L185

Added unit tests and make some minor changes to the endpoints.go file as well.

achevuru (dismissed):

pkg/resolvers/endpoints.go
```diff
@@ -185,14 +185,64 @@ func (r *defaultEndpointsResolver) resolveNetworkPeers(ctx context.Context, poli
 		} else {
 			namespaces = []string{policy.Namespace}
 		}
-		r.logger.V(1).Info("Namespaces for network peers resolution", "list", namespaces, "policy", k8s.NamespacedName(policy))
+
+		var portsToApply []policyinfo.Port
+		// populate the policy applied targets' ports
```
jayanthvn, line 190: Nit: Please add few comments as to why this is done only for ingress...

pkg/resolvers/endpoints.go
```diff
@@ -185,14 +185,64 @@ func (r *defaultEndpointsResolver) resolveNetworkPeers(ctx context.Context, poli
...
+		if policyType == networking.PolicyTypeIngress {
+			portsToApply = r.getIngressRulesPorts(ctx, policy.Namespace, &policy.Spec.PodSelector, ports)
+		}
+
 		for _, ns := range namespaces {
-			networkPeers = append(networkPeers, r.getMatchingPodAddresses(ctx, peer.PodSelector, ns, ports)...)
+			networkPeers = append(networkPeers, r.getMatchingPodAddresses(ctx, peer.PodSelector, ns, portsToApply, ports, policyType)...)
```
jayanthvn, line 196: If `portsToApply` is `nil` say for instance we hit the "Unable to List Pods" then why do even we need to go in the loop? Because eventually `portList := r.getPortList(pod, ports)` would be `0` and we just continue to the next pod isn't it? Am I missing anything?

achevuru (approved):

haouc:
go vulns check failed on go version update. we will do a follow-up to update go version in action and also build script.

PR #62, opened 2023-12-22
Enabling logs in critical paths

haouc:
**What type of PR is this?**

cleanup

Since we are running this controller in EKS managed control plane, we should enable critical logs by default.

Ran the controller locally and confirmed.
**Automation added to e2e**:

```release-note

```


jdn5126 (approved):
lgtm, thanks for improving this!

Issue #71, opened 2024-01-26
PolicyEndpoint Missing Rules with Named Ports


aballman:
```
Network Policy Controller: v1.0.4
```

I've recently switched from using Calico to the AWS CNI for enforcing network policies. I'm using ArgoCD that comes with some default network policies that reference other components by named port. I'd noticed that these policies didn't seem to be taking effect after the switch, so I dug into it and it seems that the `PolicyEndpoint` isn't being built correctly.

### What I'm providing
```
kind: NetworkPolicy
spec:
  ingress:
    - from:
        - podSelector:
            matchLabels:
              app.kubernetes.io/instance: argocd
              app.kubernetes.io/name: argocd-server
      ports:
        - port: http
          protocol: TCP
        - port: grpc
          protocol: TCP
  podSelector:
    matchLabels:
      app.kubernetes.io/instance: argocd
      app.kubernetes.io/name: argocd-dex-server
  policyTypes:
    - Ingress
```

### What it built
```
kind: PolicyEndpoint
spec:
  podIsolation:
  - Ingress
  podSelector:
    matchLabels:
      app.kubernetes.io/instance: argocd
      app.kubernetes.io/name: argocd-dex-server
  podSelectorEndpoints:
  - hostIP: 10.146.52.181
    name: argocd-dex-server-5c6dfff575-mhvzq
    namespace: argocd
    podIP: 10.146.52.212
  policyRef:
    name: argocd-dex-server
    namespace: argocd
```

### What I expected
```
kind: PolicyEndpoint
spec:
  ingress:
  - cidr: 10.146.54.136
    ports:
    - port: 5556
      protocol: TCP
    - port: 5557
      protocol: TCP
  - cidr: 10.146.59.25
    ports:
    - port: 5556
      protocol: TCP
    - port: 5557
      protocol: TCP
  podIsolation:
  - Ingress
  podSelector:
    matchLabels:
      app.kubernetes.io/instance: argocd
      app.kubernetes.io/name: argocd-dex-server
  podSelectorEndpoints:
  - hostIP: 10.146.52.181
    name: argocd-dex-server-5c6dfff575-mhvzq
    namespace: argocd
    podIP: 10.146.52.212
  policyRef:
    name: argocd-dex-server
    namespace: argocd
```

The PE is built like the above if the port is specified by number instead of by name.
The deployment and the service both have the ports named and match

haouc:
[...]

PR #72, opened 2024-01-31
ingress doesn't need to check src ports

haouc:
**What type of PR is this?**
fix a bug

bug
#71 

The controller shouldn't loop through src pods to check if their ports matching policies'.

Unit tests are updated and functional tests were done

policy applied pod (dst)
```
ports:
        - containerPort: 80
          name: http
          protocol: TCP
        - containerPort: 443
          name: https
          protocol: TCP
```
src pod
```
ports:
    - name: src
      containerPort: 9999
      protocol: TCP
```
svc
```
spec:
  ports:
  - port: 80
    name: http
    targetPort: http
  - port: 443
    name: https
    targetPort: https
```
policyendpoint
```
spec:
  ingress:
  - cidr: 192.168.xx.xxx
    ports:
    - port: 80
      protocol: TCP
    - port: 443
      protocol: TCP
  podIsolation:
  - Ingress
```

**Automation added to e2e**:

no

no

no

no
```release-note

```


pkg/resolvers/endpoints.go
```diff
@@ -327,10 +327,14 @@ func (r *defaultEndpointsResolver) getMatchingPodAddresses(ctx context.Context,
...
 		}
-		portList := r.getPortList(pod, ports)
-		if len(ports) != len(portList) && len(portList) == 0 {
-			r.logger.Info("Couldn't get matched port list from the pod", "pod", k8s.NamespacedName(&pod), "expectedPorts", ports)
-			continue
+		var portList []policyinfo.Port
+		// no need to populate src pods' ports for ingress rules
+		if rule == networking.PolicyTypeEgress {
+			portList = r.getPortList(pod, ports)
+			if len(ports) != len(portList) && len(portList) == 0 {
```
jdn5126, line 334: nit: shouldn't this be:
```
if len(portList) == 0 || len(ports) != len(portList) {
   log
   continue
}
```
since we don't want to do anything when port list is empty or port list has any mismatch

pkg/resolvers/endpoints.go
```diff
@@ -327,10 +327,14 @@ func (r *defaultEndpointsResolver) getMatchingPodAddresses(ctx context.Context,
 			r.logger.Info("pod IP not assigned yet", "pod", k8s.NamespacedName(&pod))
 			continue
 		}
-		portList := r.getPortList(pod, ports)
-		if len(ports) != len(portList) && len(portList) == 0 {
-			r.logger.Info("Couldn't get matched port list from the pod", "pod", k8s.NamespacedName(&pod), "expectedPorts", ports)
-			continue
+		var portList []policyinfo.Port
+		// no need to populate src pods' ports for ingress rules
```
achevuru, line 331: nit: Why do we even need this comment here? We're not looking in to src pods anyways

pkg/resolvers/endpoints.go
achevuru: So, we would want to proceed with the derived port list at the minimum. Say if NP specifies multiple named ports and one of them is not found in the pod spec ..we can setup probes to allow the rest. Now, if nothing is found then we don't want to allow everything for that pod IP

pkg/resolvers/endpoints.go
```diff
@@ -321,25 +335,25 @@ func (r *defaultEndpointsResolver) getMatchingPodAddresses(ctx context.Context,
...
 			r.logger.Info("pod IP not assigned yet", "pod", k8s.NamespacedName(&pod))
 			continue
 		}
-		portList := r.getPortList(pod, ports)
-		if len(ports) != len(portList) && len(portList) == 0 {
-			r.logger.Info("Couldn't get matched port list from the pod", "pod", k8s.NamespacedName(&pod), "expectedPorts", ports)
-			continue
+
+		if policyType == networking.PolicyTypeEgress {
+			portList = r.getPortList(pod, rulePorts)
```
achevuru, line 347: Why are we overwriting the `portList` here? Shouldn't we append instead? we can have the same pod selectors match on multiple unique pods/deployments. I see we're appending for the ingress but not here..

haouc: For ingress we aggregate ports from ingress rule together and apply the same set to individual pod ip. But for egress, we are setting individual entry for every pod/ip with their allowed port(s). I feel we shouldn't combine ports for egress entry.

achevuru: Makes sense, as we're adding port list per pod IP ..

achevuru (approved):

pkg/resolvers/endpoints.go
haouc: The comment was removed in latest update.

pkg/resolvers/endpoints.go
haouc: Synced offline. Changes were made to the latest update.

PR #73, opened 2024-01-31
Adding Helm chart for controller installation

jaydeokar:
**What type of PR is this?**
Scripts

N/A

This adds helm chart for the controller that can be used to install it on dataplane

N/A

Yes

**Automation added to e2e**:
N/A

N/A

N/A

N/A


charts/amazon-network-policy-controller-k8s/templates/.helmignore
jdn5126, line 1: I think this belongs in the top level of the chart, right?

charts/amazon-network-policy-controller-k8s/templates/_helpers.tpl
```diff
@@ -0,0 +1,70 @@
...
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "amazon-network-policy-controller.fullname" -}}
+{{- if .Values.fullnameOverride -}}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- $name := default .Chart.Name .Values.nameOverride -}}
+{{- if contains $name .Release.Name -}}
```
jdn5126, line 19: I see that you copied this from VPC CNI, but it feels like the `{{- end -}}` below should go before this next `if`

charts/amazon-network-policy-controller-k8s/templates/deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+          {{- end }}
+        {{- with .Values.livenessProbe }}
+        livenessProbe:
+          {{- toYaml . | nindent 10 }}
+        {{- end }}  
+        {{- with .Values.readinessProbe }}
+        readinessProbe:
+          {{- toYaml . | nindent 10 }}
+        {{- end }}
+        securityContext:
```
jdn5126, line 44: we should make `resources` configurable and consider choosing default cpu and memory request values

charts/amazon-network-policy-controller-k8s/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
```
jdn5126, line 3: nit: should we add configurable annotations here too? I was thinking in terms of ever supporting IRSA via annotations

charts/amazon-network-policy-controller-k8s/values.yaml
```diff
@@ -0,0 +1,50 @@
...
+maxConcurrentReconciles: ""
+podUpdateBatchPeriodDuration: ""
+
+securityContext:
+  allowPrivilegeEscalation: false
+  capabilities:
+    drop:
+    - ALL
+
+updateStrategy: {}
```
jdn5126, line 27: nit: why not define an update strategy by default?

charts/amazon-network-policy-controller-k8s/templates/_helpers.tpl
jaydeokar: We are pretty much using the upstream helm default template.. But shouldn't really matter the placement of 
{{- end -}}

charts/amazon-network-policy-controller-k8s/values.yaml
jaydeokar: I didn't add it as I was trying to mimic the kustomize deployment. (Even though it was not defined, it gets a default value when it is deployed in the cluster). I'll update it to the default strategy for now

charts/amazon-network-policy-controller-k8s/templates/.helmignore
jaydeokar: Good catch

charts/amazon-network-policy-controller-k8s/templates/deployment.yaml
jaydeokar: Added

charts/amazon-network-policy-controller-k8s/Chart.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v2
```
jdn5126, line 1: last nit, otherwise I think this is good to go after rebasing: shouldn't this be `v1`? I remember we had a problem with the EFA plugin helm chart where we had to go with v1: https://github.com/aws/eks-charts/blob/master/stable/aws-efa-k8s-device-plugin/Chart.yaml#L1

jdn5126 (commented):
One more comment, otherwise I can finish review after comments are addressed and PR is rebased

charts/amazon-network-policy-controller-k8s/values.yaml
```diff
@@ -0,0 +1,57 @@
...
+# Readiness probe configuration for the controller
+readinessProbe:
+  httpGet:
+    path: /readyz
+    port: 8081
+  initialDelaySeconds: 5
+  periodSeconds: 10
+
+resources:
+  limits:
```
jdn5126, line 54: I think we should avoid limits and instead just use CPU and memory requests

charts/amazon-network-policy-controller-k8s/Chart.yaml
jaydeokar: I'm curious to know the limitation we faced there. I tried deploying the helm chart in my cluster with v2, I didn't run into any issues. Even CNI migrated to v2 last year

jdn5126: It looks like CNI still uses v1: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/charts/aws-vpc-cni/Chart.yaml#L1

I vaguely recall it having to do with a helmv2 incompatibility. This may be fine here as this chart only claims to support helmv3

jaydeokar: Ah sorry, I got confused with helm v2 to v3 migration. We shouldn't be affected by these changes from v1 to v2


>Changes from v1 to v2:
>A dependencies field defining chart dependencies, which were located in a separate requirements.yaml file for v1 charts (see [Chart Dependencies](https://helm.sh/docs/topics/charts/#chart-dependencies)).
>The type field, discriminating application and library charts (see [Chart Types]
(https://helm.sh/docs/topics/charts/#chart-types)).

charts/amazon-network-policy-controller-k8s/values.yaml
jaydeokar: Added request CPU to the default values

jdn5126 (approved):

PR #80, opened 2024-02-02
combine endpoints based on cidr

haouc:
**What type of PR is this?**

improvement

We should combine same cidr's entries together to avoid node agent expensive operations.

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/policyendpoints/manager.go
+++ b/pkg/policyendpoints/manager.go
@@ -181,7 +181,7 @@ func (m *policyEndpointsManager) processPolicyEndpoints(pes []policyinfo.PolicyE
 	return newPEs
 }
 
-// the controller should consolidate the ingress endpoints and put entries to one CIDR if they belong to a same cidr
+// the controller should consolidate the ingress and egress endpoints and put entries to one CIDR if they belong to a same CIDR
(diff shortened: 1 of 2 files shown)


pkg/policyendpoints/manager.go
```diff
@@ -167,7 +167,37 @@ func (m *policyEndpointsManager) computePolicyEndpoints(policy *networking.Netwo
...
+	for _, pe := range pes {
+		pe.Spec.Ingress = combineRulesEndpoints(pe.Spec.Ingress)
+		pe.Spec.Egress = combineRulesEndpoints(pe.Spec.Egress)
+		newPEs = append(newPEs, pe)
+	}
+	m.logger.Info("manager processed policy endpoints to consolidate rules", "preLen", len(pes), "postLen", len(newPEs), "newPEs", newPEs)
+	return newPEs
+}
+
+// the controller should consolidate the ingress endpoints and put entries to one CIDR if they belong to a same cidr
```
achevuru, line 184: Why ingress only endpoints? Should be same for egress endpoints as well, isn't it..

haouc: Yeah, the comment was not updated. They both are deduped. I will update the comment.

achevuru (approved):

PR #89, opened 2024-03-15
ignore host network pods in all endpoints

haouc:
**What type of PR is this?**

bug
We shouldn't populate pods in host network to PolicyEndpoint's endpoints.

The endpoints from host network pods in ingress/egress/podselector can become redundant data in nodes and CRs. 

Tested in local dev cluster and cyclonus tests were passed.

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/aws/amazon-network-policy-controller-k8s
 
-go 1.21
+go 1.22.0
+
 toolchain go1.22.2
 
 require (


achevuru:
Let's add a simple test result with host networking pods matching ingress, egress and pod selector endpoint selectors..

pkg/resolvers/endpoints.go
```diff
@@ -491,3 +497,22 @@ func dedupPorts(policyPorts []policyinfo.Port) []policyinfo.Port {
 	}
 	return nil
 }
+
+func (r *defaultEndpointsResolver) getPodList(ctx context.Context, opts client.ListOption) ([]corev1.Pod, error) {
```
jayanthvn, line 501: Maybe to make this more clear, it can be `getNonHostNetworkingPodList` ?

haouc:
> Let's add a simple test result with host networking pods matching ingress, egress and pod selector endpoint selectors..

The results doc has been shared offline. Please let me know if we want to add the result here as well.

pkg/resolvers/endpoints.go
haouc: updated the name.

PR #92, opened 2024-03-27
Allow named ports for IPBlocks on ingress

jchen6585:
**What type of PR is this?**

bug

#81 

Allows named ports when using IPBlocks (ingress rules only)

Manually applied Cx config and policyendpoint looks fine
```
Spec:
  Ingress:
    Cidr:  172.17.0.0/16
    Ports:
      Port:      80
      Protocol:  TCP
    Cidr:        192.168.8.106
    Ports:
      Port:      443
      Protocol:  TCP
  Pod Isolation:
    Ingress
  Pod Selector:
    Match Labels:
      App:  target
  Pod Selector Endpoints:
    Host IP:    192.168.98.89
    Name:       target
    Namespace:  default
    Pod IP:     192.168.125.203
  Policy Ref:
    Name:       allow-web-traffic
    Namespace:  default
```

Using allow all in ingress rule
```
Spec:
  Ingress:
    Cidr:  0.0.0.0/0
    Ports:
      Port:      80
      Protocol:  TCP
      Port:      443
      Protocol:  TCP
    Cidr:        ::/0
    Ports:
      Port:      80
      Protocol:  TCP
      Port:      443
      Protocol:  TCP
  Pod Isolation:
    Ingress
  Pod Selector:
    Match Labels:
      App:  target
  Pod Selector Endpoints:
    Host IP:    192.168.98.89
    Name:       target
    Namespace:  default
    Pod IP:     192.168.116.203
  Policy Ref:
    Name:       allow-web-traffic
    Namespace:  default
```

**Automation added to e2e**:

1. NP allowing ingress from IPBlock + 2 named ports
2. NP allowing ingress from all CIDRs + 2 named ports

No

No

Yes

```release-note
Allow for using named ports when using IPBlocks for the ingress source.
```

### DIFF
--- a/pkg/resolvers/endpoints.go
+++ b/pkg/resolvers/endpoints.go
@@ -65,7 +65,7 @@ func (r *defaultEndpointsResolver) computeIngressEndpoints(ctx context.Context, 
 	for _, rule := range policy.Spec.Ingress {
 		r.logger.V(1).Info("computing ingress addresses", "peers", rule.From)
 		if rule.From == nil {
-			ingressEndpoints = append(ingressEndpoints, r.getAllowAllNetworkPeers(rule.Ports)...)
+			ingressEndpoints = append(ingressEndpoints, r.getAllowAllNetworkPeers(ctx, policy, rule.Ports, networking.PolicyTypeIngress)...)
 			continue
 		}
 		resolvedPeers, err := r.resolveNetworkPeers(ctx, policy, rule.From, rule.Ports, networking.PolicyTypeIngress)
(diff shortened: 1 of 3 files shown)


pkg/resolvers/endpoints.go
```diff
@@ -163,8 +169,12 @@ func (r *defaultEndpointsResolver) resolveNetworkPeers(ctx context.Context, poli
...
 			var portList []policyinfo.Port
 			for _, port := range ports {
-				if port := r.convertToPolicyInfoPortForCIDRs(port); port != nil {
-					portList = append(portList, *port)
+				portInfo := r.convertToPolicyInfoPortForCIDRs(port)
+				if portInfo != nil {
+					portList = append(portList, *portInfo)
+				} else {
+					ports := r.getIngressRulesPorts(ctx, policy.Namespace, &policy.Spec.PodSelector, []networking.NetworkPolicyPort{port})
+					portList = append(portList, ports...)
```
haouc, line 177: We should only support named ports for ingress. I think we need do same if condition check on `policyType` here as line 140.

haouc (approved):
lgtm

PR #99, opened 2024-04-11
Update manager_auth_proxy_patch.yaml

orsenthil:
Update the reference from gcr.io to registry.k8s.io

> kube-rbac-proxy is moving to  registry.k8s.io/kubebuilder/kube-rbac-proxy  (from gcr.io/kubebuilder/kube-rbac-proxy)  because GCR is being sunset. We need to update these references.


config/default/manager_auth_proxy_patch.yaml
```diff
@@ -31,7 +31,7 @@ spec:
           capabilities:
             drop:
               - "ALL"
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.13.1
+        image: registry.k8s.io/kubebuilder/kube-rbac-proxy:v0.13.1
```
jaydeokar, line 34: These images are already present in the upstream registry ?

jaydeokar (approved):
lgtm.. Just one comment to verify if images are present in upstream

orsenthil:
Yes, the image is present - https://explore.ggcr.dev/?repo=registry.k8s.io%2Fkubebuilder%2Fkube-rbac-proxy

PR #101, opened 2024-04-11
Add docker-build target for controller build

jaydeokar:
**What type of PR is this?**
cleanup

N/A

This adds a docker build target for internal build systems

N/A

Yes

**Automation added to e2e**:
N/A

N/A

N/A

N/A


Makefile
```diff
@@ -176,12 +176,18 @@ GO_IMAGE_TAG=$(shell cat .go-version)
 BUILD_IMAGE=public.ecr.aws/docker/library/golang:$(GO_IMAGE_TAG)
 BASE_IMAGE=public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-nonroot:latest.2
 GO_RUNNER_IMAGE=public.ecr.aws/eks-distro/kubernetes/go-runner:v0.15.0-eks-1-27-3
-.PHONY: docker-buildx
-docker-buildx: test
+.PHONY: docker-buildx-and-test
+docker-buildx-and-test: test
```
haouc, line 180: I think there is no need to create another target just for different name. Adding test to docker-buildx is fine for me.

.go-version
```diff
@@ -1 +1 @@
-1.22.2
+1.22
```
jaydeokar, line 1: This pins the go-version to the latest golang language version i.e. `v1.N` instead of the patch version `v1.N.P`.. 
This makes sure we are always building image with the latest patch for the golang version and does not need regular updates to pick the right patch version.

jayanthvn (approved):

PR #177, opened 2025-05-16
add flag to paginate list calls

oliviassss:
**What type of PR is this?**

When the pod scale is large, the current unpaginated list pod call will heavy load apiserver and lead to apiserver timeout.
Add a flag `--list-page-size` to paginate k8s list pod calls, default value is `1000`.

Avoid apiserver timeout issue that may lead to controller crash when pod scale is large.

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -102,6 +102,7 @@ func main() {
 	setupLog.Info("Checking args for PE chunk size", "PEChunkSize", controllerCFG.EndpointChunkSize)
 	setupLog.Info("Checking args for policy batch time", "NPBatchTime", controllerCFG.PodUpdateBatchPeriodDuration)
 	setupLog.Info("Checking args for reconciler count", "ReconcilerCount", controllerCFG.MaxConcurrentReconciles)
+	setupLog.Info("Checking args for k8s list call page size", "ListPageSize", controllerCFG.ListPageSize)
 
 	if controllerCFG.EnableConfigMapCheck {
 		var cancelFn context.CancelFunc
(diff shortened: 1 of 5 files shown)


M00nF1sh (approved):
lgtm for this code alone，but i have below concerns:
1. **have we tested that this pagination is really happening in apiServer from audit logs?** are we use real client instead of cached one? 
   * edit: just read through the entire codebase, from the code seems we are using the cached client, i'm wondering how this would work at all..
2.  i'm curious that there isn't pagination support in client-go or controller-runtime?

M00nF1sh (changes requested):
k8sClient shall be a real client to make this work, and seems we are using cached one, unless i made some mistake here.
https://github.com/aws/amazon-network-policy-controller-k8s/blob/main/pkg/config/runtime_config.go#L96

oliviassss:
@M00nF1sh thanks, yes it's a cached client. I'm running the controller in a cluster with 130k pods to confirm from apiserver side.

oliviassss:
>i'm curious that there isn't pagination support in client-go or controller-runtime

`client-go` enforces a default page size as 500, but for init list call, with RV=0, the limit is ignored. so apiserver ignores page size when RV=0. saw this from source code comment
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/cacher/cacher.go#L693

```
// computeListLimit determines whether the cacher should
// apply a limit to an incoming LIST request and returns its value.
//
// note that this function doesn't check RVM nor the Continuation token.
// these parameters are validated by the shouldDelegateList function.
//
// as of today, the limit is ignored for requests that set RV == 0
func computeListLimit(opts storage.ListOptions) int64 {
	if opts.Predicate.Limit <= 0 || opts.ResourceVersion == "0" {
		return 0
	}
	return opts.Predicate.Limit
}
```
I confirmed this behavior from 130k-pod cluster
1. During startup only 1 list pods call is made (although the limit=500, I don't see pagination)
2. During reconciliation later, only watch calls are mede, no list calls (should because it's list from a cached client). Also, in reconciliation the list pod calls have selector, so it's not listing all pods, which is less concerned. I will do more scale test to confirm.

oliviassss:
Closed as paginated list from APIserver cache is not supported before 1.34

PR #180, opened 2025-06-05
Add watcher for CRD to restart controller

Issacwww:
**What type of PR is this?**
feature

N/A

Currently the PolicyEndpoint CRD is part of vpc-cni addon, the addon got removed the CRD will be uninstalled.

Add the crd manager to ensure this CRD be always available

Scenario 1
- Create cluster with VPC-CNI addon -> PolicyEnpoint CRD installed originally
```
k get crds  | grep policy
policyendpoints.networking.k8s.aws           2025-06-18T20:43:14Z
```
- Manual remove CRD
```
k delete crd policyendpoints.networking.k8s.aws
customresourcedefinition.apiextensions.k8s.io "policyendpoints.networking.k8s.aws" deleted
```
- The crd manager restart the controller and recreate it 
```
{"level":"info","ts":"2025-06-18T20:51:28Z","logger":"setup.crd-manager","msg":"CRD doesn't exist, creating it","name":"policyendpoints.networking.k8s.aws"}
```
```
k get crds  | grep policy
policyendpoints.networking.k8s.aws           2025-06-18T20:51:28Z
```

Scenario 2
- Remove vpc-cni via delete-addon call
```
beks delete-addon --cluster-name npc-crd --addon-name vpc-cni
{
    "addon": {
        "addonName": "vpc-cni",
        "clusterName": "npc-crd",
        "status": "DELETING",
        "addonVersion": "v1.19.5-eksbuild.1",
        "health": {
            "issues": []
        },
        "addonArn": "arn:aws:eks:us-west-2:xxx:addon/npc-crd/vpc-cni/44cbc260-bad5-3893-6384-099d7d725bc0",
        "createdAt": "2025-06-18T13:36:02.120000-07:00",
        "modifiedAt": "2025-06-18T13:53:52.015000-07:00",
        "tags": {}
    }
}
```
- crd still represent
```
k get crds  | grep policy
policyendpoints.networking.k8s.aws           2025-06-18T20:51:28Z
```

Scenario 3
- Reinstall vpc-cni via create-addon call (with the short-term fix for the annotation mismatch)
```
beks describe-addon --cluster-name npc-crd --addon-name vpc-cni
{
    "addon": {
        "addonName": "vpc-cni",
        "clusterName": "npc-crd",
        "status": "ACTIVE",
        "addonVersion": "v1.19.5-eksbuild.1",
        "health": {
            "issues": []
        },
        "addonArn": "arn:aws:eks:us-west-2:171391670848:addon/npc-crd/vpc-cni/58cbc26a-43db-5d07-d5f3-2b723ce148db",
        "createdAt": "2025-06-18T13:56:50.544000-07:00",
        "modifiedAt": "2025-06-18T13:57:29.607000-07:00",
        "tags": {}
    }
}
```
it pass and do not recreate the crd
```
k get crds  | grep policy
policyendpoints.networking.k8s.aws           2025-06-18T20:51:28Z
```

**Automation added to e2e**:

see `crd_manager_test.go`

N/A

should not break upgrades or downgrades.

```release-note

```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,12 @@ COPY api/ api/
 COPY pkg/ pkg/
 COPY internal/ internal/
 
+# Copy CRD file for embedding
+COPY charts/amazon-network-policy-controller-k8s/crds/crds.yaml pkg/crd/crds.yaml
+# Short-term workaround for controller-gen version mismatch, should remove once the CRD removed from vpc-cni addon
+RUN sed -i 's/controller-gen.kubebuilder.io\/version: v[0-9]\+\.[0-9]\+\.[0-9]\+/controller-gen.kubebuilder.io\/version: v0.11.3/' pkg/crd/crds.yaml
+
+
 # Version package for passing the ldflags
 # TODO: change this to network controller's version
 ENV VERSION_PKG=github.com/aws/amazon-network-policy-controller-k8s/pkg/version
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -42,6 +42,7 @@ import (
 	policyinfo "github.com/aws/amazon-network-policy-controller-k8s/api/v1alpha1"
 	"github.com/aws/amazon-network-policy-controller-k8s/internal/controllers"
 	"github.com/aws/amazon-network-policy-controller-k8s/pkg/config"
+	"github.com/aws/amazon-network-policy-controller-k8s/pkg/crd"
 	"github.com/aws/amazon-network-policy-controller-k8s/pkg/k8s"
 	"github.com/aws/amazon-network-policy-controller-k8s/pkg/policyendpoints"
 	"github.com/aws/amazon-network-policy-controller-k8s/pkg/utils/configmap"
@@ -103,9 +104,9 @@ func main() {
 	setupLog.Info("Checking args for policy batch time", "NPBatchTime", controllerCFG.PodUpdateBatchPeriodDuration)
 	setupLog.Info("Checking args for reconciler count", "ReconcilerCount", controllerCFG.MaxConcurrentReconciles)
 
+	var cancelFn context.CancelFunc
+	ctx, cancelFn = context.WithCancel(ctx)
 	if controllerCFG.EnableConfigMapCheck {
-		var cancelFn context.CancelFunc
-		ctx, cancelFn = context.WithCancel(ctx)
 		setupLog.Info("Enable network policy controller based on configuration", "configmap", configmap.GetControllerConfigMapId())
 		configMapManager := config.NewConfigmapManager(configmap.GetControllerConfigMapId(),
 			clientSet, cancelFn, configmap.GetConfigmapCheckFn(), ctrl.Log.WithName("configmap-manager"))
@@ -120,6 +121,11 @@ func main() {
 		}
 	}
 
+	crdManager := crd.NewCRDManager(restCFG, setupLog.WithName("crd-manager"), cancelFn)
+	if err := crdManager.Start(ctx); err != nil {
+		setupLog.Error(err, "Unable to ensure CRDs exist")
+		os.Exit(1)
+	}
 	rtOpts := config.BuildRuntimeOptions(controllerCFG.RuntimeConfig, scheme)
 
 	mgr, err := ctrl.NewManager(restCFG, rtOpts)
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -50,6 +50,17 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - patch
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -15,6 +15,15 @@ rules:
   - list
   - watch
 - apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  verbs:
+  - create
+  - get
+  - list
+  - watch
+- apiGroups:
   - networking.k8s.aws
   resources:
   - policyendpoints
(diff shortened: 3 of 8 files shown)


Dockerfile
```diff
@@ -20,6 +20,7 @@ COPY cmd/main.go cmd/main.go
 COPY api/ api/
 COPY pkg/ pkg/
 COPY internal/ internal/
+COPY config/crd/bases/ internal/controllers/
```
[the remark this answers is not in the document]
Issacwww: this seems have problem in the github CI task.
consider alternatives
- copy and commit the file as part of codegen (recommended)
- use inline yaml within crd_controller

cmd/main.go
```diff
@@ -133,12 +135,18 @@ func main() {
...
 	policyController := controllers.NewPolicyReconciler(mgr.GetClient(), policyEndpointsManager,
 		controllerCFG, finalizerManager, ctrl.Log.WithName("controllers").WithName("policy"))
+	crdController := controllers.NewCRDReconciler(mgr.GetClient(), ctrl.Log.WithName("controllers").WithName("crd"))
 	if enableNetworkPolicyController {
 		setupLog.Info("Network Policy controller is enabled, starting watches")
 		if err := policyController.SetupWithManager(ctx, mgr); err != nil {
 			setupLog.Error(err, "Unable to setup network policy controller")
 			os.Exit(1)
 		}
+		setupLog.Info("starting watch policy endpoint crd...")
```
Issacwww, line 145: @yash97  want to double confirm this, given CRD will be created when cluster bootstrap, we should be good to move this to outside of the `enableNetworkPolicyController` condition.

yash97: yes. Creation of policy endpoint crd is independent of config npc to true

jaydeokar: There could be some specific use case where the user uninstalls the CRD in their cluster as they don't use NPC. 
If we move it outside, the CRD will be created always even if the controller is not enabled. I am not sure we would want to have that behavior

internal/controllers/crd_controller.go
```diff
@@ -0,0 +1,96 @@
...
+// +kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list;watch;create;update
+
+func (r *CRDReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	_ = logf.FromContext(ctx)
+	r.logger.Info("Got reconcile request", "resource", req)
+	if req.Name != policyendpointsCrdName {
+		r.logger.Info("Ignoring reconcile request for non-policyendpoints CRD", "name", req.Name)
+		return ctrl.Result{}, nil
+	}
+	desiredCRD, err := decodePolicyEndpointsCrd()
```
viveksb007, line 74: do we need to decode this for every `Reconcile()` call? or we can do once while creating CRDReconciler object?

looks like a static one time thing. we can probably fail fast if `crds.yaml` is not found. Also i think we can reference `policyendpointsCrdName` after decoding the CRD only.

internal/controllers/crds.yaml
viveksb007, line 1: is there going to be any difference b/w this crds.yaml and https://github.com/aws/amazon-network-policy-controller-k8s/blob/main/config/crd/bases/networking.k8s.aws_policyendpoints.yaml?

if not, is there someway to re-use that?

internal/controllers/suite_test.go
```diff
@@ -65,12 +67,23 @@ var _ = BeforeSuite(func() {
...
 	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
 	Expect(err).NotTo(HaveOccurred())
 	Expect(k8sClient).NotTo(BeNil())
 
+	mgr, err := ctrl.NewManager(cfg, ctrl.Options{Scheme: scheme.Scheme})
+	Expect(err).NotTo(HaveOccurred())
+
+	err = (&CRDReconciler{
+		k8sClient: k8sClient,
+		logger:    logf.Log.WithName("controllers").WithName("crd"),
```
yash97, line 83: Nit policyendpoint-crd could be better.

internal/controllers/crds.yaml
Issacwww: There is online differ, the annotation version. There is a sed command to update it as this version mismatching causing the crd failed to be updated when recreate vpc-cni addon that as the crd with annoation `v0.11.3`

The sed update will be removed once we that crd removed from addon manifest. 

However, even after that, we can no re-use the existing yaml, and the go:embed has the limitation to use the file in parent folder.  see https://github.com/golang/go/issues/46056

internal/controllers/crd_controller.go
Issacwww: will update to skip this every time

internal/controllers/crd_controller.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to load CRD from file: %w", err)
+	}
+	existing := &apiextensionsv1.CustomResourceDefinition{}
+	err = r.k8sClient.Get(ctx, types.NamespacedName{Name: desiredCRD.Name}, existing)
+	if err != nil {
+		if client.IgnoreNotFound(err) == nil {
+			r.logger.Info("CRD not found, creating...")
+			return ctrl.Result{}, r.k8sClient.Create(ctx, desiredCRD)
```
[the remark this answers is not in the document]
Issacwww: good catch, will add it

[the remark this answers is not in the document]
Issacwww: Will add metric later as we need to have custom health server to handle config map issue as well

Dockerfile
```diff
@@ -21,6 +21,12 @@ COPY api/ api/
 COPY pkg/ pkg/
 COPY internal/ internal/
 
+# Copy CRD file for embedding
```
haouc, line 24: You can use `go:embed` decoration to let compiler build them into a []byte variable.

config/rbac/role.yaml
```diff
@@ -14,6 +14,18 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  verbs:
+  - create
```
haouc, line 22: I think we only need create/get/list if we install on start. Or list/watch/create if we want to reconcile and reinstall.

Dockerfile
Issacwww: yes, I used go:embed, but the limitation of with go:embed is it cannot read from the parent folder. 
https://github.com/golang/go/issues/46056

and i use this copy as an automation that we only need to manage the source yaml.

yash97:
lgtm 🚀

yash97 (approved):

PR #191, opened 2025-09-16
Only combine Rules Endpoints with the same CIDR if they share identical Exception lists

Subzidion:
See #193


Subzidion:
See #193

PR #193, opened 2025-09-18
Only combine Rules Endpoints with the same CIDR if they share identical Exception lists

Subzidion:
Only combine Rules Endpoints with the same CIDR if they share identical Exception lists. Currently, Rules Endpoints are combined if the share the same CIDR, however, some of these Endpoints are incompatible when combined with each other. 

The example from #146 shows a `NetworkPolicy` with 2 Egress Rules, one allowing egress out to 0.0.0.0/0 _except_ to CIDRs `10.0.0.0/8`, `172.16.0.0/12`, and `192.168.0.0/16`. The second rule allowing _all_ egress traffic out for TCP and UDP on Port 53. Both rules egress out `0.0.0.0/0` and so the controller merges their internal representation. However, this is contradictory, as the second rule _should_ allow Port 53 (DNS) traffic to any CIDR, including the ones mentioned in the `except` list of Rule 1.

**Note**: This undoes a lot of #80, so we need to consider the performance impact of making this operation more expensive by doing this.

Rule:
```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: curl
  namespace: default
spec:
  egress:
  - to:
    - ipBlock:
        cidr: 0.0.0.0/0
        except:
          - 10.0.0.0/8
          - 172.16.0.0/12
          - 192.168.0.0/16
  - ports:
      - protocol: UDP
        port: 53
      - port: 53
        protocol: TCP
  policyTypes:
    - Egress
  podSelector:
    matchLabels:
      app: curl
```

This is again visible from #146 in the `PolicyEndpoint` where the two separate rules appear merged under the single 0.0.0.0/0 Egress Rule.

```
Name:         curl-4mhvr
Namespace:    test1
Labels:       <none>
Annotations:  <none>
API Version:  networking.k8s.aws/v1alpha1
Kind:         PolicyEndpoint
Metadata:
  Creation Timestamp:  2024-11-12T23:06:46Z
  Generate Name:       curl-
  Generation:          1
  Owner References:
    API Version:           networking.k8s.io/v1
    Block Owner Deletion:  true
    Controller:            true
    Kind:                  NetworkPolicy
    Name:                  curl
    UID:                   acc0d91d-e6ab-46fe-a8bb-aad64e6eaaec
  Resource Version:        1073029
  UID:                     7415e7fe-6711-4df3-8664-ba5ae73d71a1
Spec:
  Egress:
    Cidr:  ::/0
    Ports:
      Port:      53
[...]

PR #205, opened 2026-01-15
Fix combineRulesEndpoints to correctly handle allow-all port semantics

### LINKED ISSUE #197: Policy endpoint does not accurately reflect the network policy
I had opened https://github.com/aws/aws-network-policy-agent/issues/481 after we experienced a change in behavior since upgrading to a version of the agent newer than 2.3.1 however after further investigation it looks like the potential issue could lie within the policy controller and the way in which is creates policy endpoint for network policies.

As an example, given the following network policy:

```
kind: NetworkPolicy
metadata:
  annotations:
  name: test
spec:
  egress:
  - ports:
    - port: 53
      protocol: UDP
    - port: 53
      protocol: TCP
    to:
    - ipBlock:
        cidr: 172.16.0.0/12
  - to:
    - ipBlock:
        cidr: 0.0.0.0/0
        except:
        - 172.16.0.0/12
  - to:
    - ipBlock:
        cidr: 172.16.0.0/12
  podSelector:
    matchLabels:
      app: demo
  policyTypes:
  - Egress
```

We see the following policy endpoint created:

```
kind: PolicyEndpoint
metadata:
  name: test-csl5v
spec:
  egress:
  - cidr: 172.16.0.0/12
    ports:
    - port: 53
      protocol: UDP
    - port: 53
      protocol: TCP
  - cidr: 0.0.0.0/0
    except:
    - 172.16.0.0/12
  podIsolation:
  - Egress
  podSelector:
    matchLabels:
      app: demo
  policyRef:
    name: test
    namespace: test
```

The ANY egress rule on `172.16.0.0/12` that exists on the network policy is not present in the policy endpoint, is this expected? I would expect the policy endpoint to reflect the network policy and include this egress rule.

Supreeth095:
When combining rules for the same CIDR, if any rule allows all ports (empty Ports slice), the combined rule should allow all ports.

Previously, the function would incorrectly append the empty slice, resulting in policies being more restrictive than intended.

Fixes #197 #193

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/policyendpoints/manager.go
+++ b/pkg/policyendpoints/manager.go
@@ -291,7 +291,17 @@ func combineRulesEndpoints(ingressEndpoints []policyinfo.EndpointInfo) []policyi
 	for _, iep := range ingressEndpoints {
 		if _, ok := combinedMap[string(iep.CIDR)]; ok {
 			tempIEP := combinedMap[string(iep.CIDR)]
-			tempIEP.Ports = append(combinedMap[string(iep.CIDR)].Ports, iep.Ports...)
+			// Handle port semantics correctly:
+			// - Empty Ports slice means "allow all ports"
+			// - If any rule allows all ports, the combined rule should allow all ports
+			if len(iep.Ports) == 0 {
+				// New rule allows all ports, so combined rule should allow all ports
+				tempIEP.Ports = []policyinfo.Port{}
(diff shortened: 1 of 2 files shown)


stefansedich:
Hi @Supreeth095 do we have any idea on when this change might be merged and deployed to the controlplane?

PR #206, opened 2026-01-16
Exclude hostNetwork pods from PolicyEndpoints

### LINKED ISSUE #202: Network Policy results in PolicyEndpoints which have hostNetwork pods
Currently if we have a `hostNetwork` pod which satisfies the condition for PodSelection in Network Policy, then that pod is included in PolicyEndpoints.

Example Network Policy
```
kind: NetworkPolicy
metadata:
  name: allow-all-egress
spec:
  podSelector: {}
  egress:
    - {}
  policyTypes:
    - Egress

---
apiVersion: v1
kind: Pod
metadata:
  name: dnsutils-test-pod
  namespace: default
spec:
  hostNetwork: true
  containers:
  - name: dnsutils
    image: registry.k8s.io/e2e-test-images/agnhost:2.39
    command:
      - sleep
      - "3600"
    imagePullPolicy: IfNotPresent
  restartPolicy: Always
  
```

```
Resulting NP and PE

kind: NetworkPolicy
spec:
  egress:
  - {}
  podSelector: {}
  policyTypes:
  - Egress

kind: PolicyEndpoint
spec:
  egress:
  - cidr: 0.0.0.0/0
  - cidr: ::/0
  podIsolation:
  - Egress
  podSelector: {}
  podSelectorEndpoints:
  - hostIP: 192.168.140.47
    name: dnsutils-test-pod
    namespace: default
    podIP: 192.168.140.47
  - hostIP: 192.168.140.47
    name: node-debugger-i-04d5987cc9d2b6d27-twv6x
    namespace: default
    podIP: 192.168.140.47
  policyRef:
    name: allow-all-egress
    namespace: default


```

error logs in Network Policy Agent
```
{"level":"debug","ts":"2025-12-10T06:33:02.101Z","caller":"controllers/policyendpoints_controller.go:295","msg":"Got the podIdentifierLock for Pod: dnsutils-test-pod, Namespace: default, PodIdentifier: dnsutils-test-default"}
{"level":"error","ts":"2025-12-10T06:33:02.101Z","caller":"ebpf/bpf_client.go:785","msg":"Not found any interface starting with prefixes and the hash. Prefixes searched [eni vlan] hash ccfcaa8937a error 2 errors occurred:\n\t* failed to find link eniccfcaa8937a: Link not found\n\t* failed to find link vlanccfcaa8937a: Link not found\n\n"}
{"level":"warn","ts":"2025-12-10T06:33:02.101Z","caller":"controllers/policyendpoints_controller.go:295","msg":"Failed to attach ebpf probes for pod dnsutils-test-pod in namespace default. Pod might have been deleted"}
{"level":"error","ts":"2025-12-10T06:33:02.101Z","caller":"controllers/policyendpoints_controller.go:263","msg":"Failed to attach eBPF probes for pod dnsutils-test-pod namespace default : 2 errors occurred:\n\t* failed to find link eniccfcaa8937a: Link not found\n\t* failed to find link vlanccfcaa8937a: Link not found\n\n"}
{"level":"error","ts":"2025-12-10T06:33:02.101Z","caller":"controllers/policyendpoints_controller.go:145","msg":"Error configuring eBPF Probes 2 errors occurred:\n\t* failed to find link eniccfcaa8937a: Link not found\n\t* failed to find link vlanccfcaa8937a: Link not found\n\n"}
```

> Solution (need to add both)
1. NPC should not include the host networking pods in the resulting PolicyEndpoint
2. While reconciling, PE controller in Network Policy Agent (https://github.com/aws/aws-network-policy-agent) should have a check on pod selectors to ignore the pods if PodIP is equal to HostIP.

emilyhuaa:
**What type of PR is this?**

bug

Fixes #202

This PR excludes hostNetwork pods from PolicyEndpoint resources. Currently, the NPC includes hostNetwork pods when they match a NetworkPolicy's podSelector, which causes errors in the Network Policy Agent because hostNetwork pods share the node's network namespace and don't have dedicated ENI or VLAN interfaces that eBPF probes can attach to.

- Added unit tests for `IsHostNetworkPod` helper function
- Added integration test `TestEndpointsResolver_ExcludesHostNetworkPods_Integration` that validates:
  - NetworkPolicy with empty podSelector correctly excludes hostNetwork pods
  - Mixed scenarios with both hostNetwork and non-hostNetwork pods work correctly
  - Only non-hostNetwork pods appear in PolicyEndpoint resources
- All existing unit tests pass
- Verified hostNetwork filtering works across all three policy types (NetworkPolicy, ApplicationNetworkPolicy, ClusterNetworkPolicy)

**Automation added to e2e**:

Integration test added: `TestEndpointsResolver_ExcludesHostNetworkPods_Integration` in `pkg/resolvers/endpoints_test.go`

Np

No breaking changes. This is a filtering enhancement that only affects which pods are included in PolicyEndpoints. Existing non-hostNetwork pods continue to work as before. HostNetwork pods were previously incorrectly included and causing errors, so this fix improves behavior.

Yes

```release-note
Exclude hostNetwork pods from PolicyEndpoint resources to prevent eBPF probe attachment errors. HostNetwork pods now correctly bypass network policy enforcement since they use the node's network namespace.
```


pkg/resolvers/endpoints.go
```diff
@@ -352,6 +359,13 @@ func (r *defaultEndpointsResolver) getMatchingPodAddresses(ctx context.Context,
 	r.logger.V(1).Info("Got pods for label selector", "count", len(podList.Items), "selector", ls.String())
 
 	for _, pod := range podList.Items {
+		// Skip hostNetwork pods
```
viveksb007: we need to only remove pods from `PodSelectorEndpoints`,  it seems this part is removing from ingress/egress rules. can you validate this once?

pkg/k8s/pod_utils.go
```diff
@@ -27,6 +27,11 @@ func IsPodNetworkReady(pod *corev1.Pod) bool {
 		pod.Status.Phase != corev1.PodFailed
 }
 
+// IsHostNetworkPod returns true if the pod uses the host network namespace
+func IsHostNetworkPod(pod *corev1.Pod) bool {
```
viveksb007: we don't need to extract this out in a function as its simple field reference in `pod` object

pkg/resolvers/endpoints_test.go
```diff
@@ -1176,3 +1178,119 @@ func TestEndpointsResolver_ExcludesTerminalPods(t *testing.T) {
...
+			},
+			PolicyTypes: []networking.PolicyType{
+				networking.PolicyTypeIngress,
+				networking.PolicyTypeEgress,
+			},
+		},
+	}
+
+	// Resolve endpoints
+	_, _, podEndpoints, err := resolver.Resolve(context.Background(), policy)
```
viveksb007, line 1261: please add a validation in ingress and egress endpoints also. Lets add a test validating that hostNetworking pods are present in ingress/egress endpoints and are only skipped in case of PodEndpointSelectors

viveksb007 (dismissed):

pkg/resolvers/endpoints.go
emilyhuaa: Fixed! Removed hostNetwork filtering from getMatchingPodAddresses() so they're only excluded from PodSelectorEndpoints. Added test to verify.

pkg/resolvers/endpoints.go
```diff
@@ -116,6 +116,12 @@ func (r *defaultEndpointsResolver) computePodSelectorEndpoints(ctx context.Conte
 		return nil, err
 	}
 	for _, pod := range podList.Items {
+		if pod.Spec.HostNetwork {
+			r.logger.V(1).Info("Skipping hostNetwork pod",
```
oliviassss: nit: we should remove the V(1) here, as the prod logs will only show V(0) as default setting.

emilyhuaa: All the other logs in the same file use `r.logger.V(1).Info` so I maintained this for consistency. Should I still remove the V(1)?

oliviassss: Yes, let's remove it to improve observability. Users may have hostNetwork pods targeted by their netpol. This change will aid debugging by ensuring the controller doesn't silently skip these pods.

emilyhuaa: Ok got it, just removed the V(1) from that line. Thank you!

pkg/resolvers/endpoints.go
```diff
@@ -116,6 +116,12 @@ func (r *defaultEndpointsResolver) computePodSelectorEndpoints(ctx context.Conte
 		return nil, err
 	}
 	for _, pod := range podList.Items {
+		if pod.Spec.HostNetwork {
+			r.logger.Info("Skipping hostNetwork pod",
```
oliviassss: to reduce log noise, can we batch the logs in one. maintain a list of `skippedHostNetworkPods`, when this for loop ends, and the list is not empty. we log out something like 
```
"Skipped hostNetwork pods for policy %s: ..."
```

oliviassss (dismissed):

oliviassss:
lgtm

oliviassss (approved):

PR #208, opened 2026-02-02
added processing duration around Reconcile method

viveksb007:
**What type of PR is this?**
- debugging improvement
- go version update

we are adding latency numbers for the specific resource being reconciled. This will help debug the Reconciliations which exhibit latency 

**Automation added to e2e**:

**Does this PR introduce any user-facing change?**: NO

```release-note

```


yash97:
do you think we should also export this as metric?

viveksb007:
there is already in-direct metric emitted corresponding to this `controller_runtime_reconcile_time_seconds_network_policy`, but that metric doesn't tell which resource exhibited the latency. This log will help in actually chasing the NetworkPolicy resource which is resulting in latency

yash97 (dismissed):

internal/controllers/policy_controller.go
```diff
@@ -98,8 +98,11 @@ type policyReconciler struct {
 //+kubebuilder:rbac:groups="networking.k8s.io",resources=networkpolicies,verbs=get;list;watch;update;patch
 
 func (r *policyReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	startTime := time.Now()
 	r.logger.Info("Got reconcile request", "resource", request)
-	return ctrl.Result{}, r.reconcile(ctx, request)
+	err := r.reconcile(ctx, request)
+	r.logger.Info("Reconcile completed", "resource", request, "duration", time.Since(startTime))
```
parlakisik: You can also share whether reconcile was successful or not. This can help you to evaluate the issue in the same log

viveksb007: yeah good idea

jaydeokar (approved):
lgtm

PR #210, opened 2026-02-13
fixing Service named port bypass issue

viveksb007:
**What type of PR is this?**
bugfix

Fixes a security bypass in PolicyEndpoint generation where Service ClusterIPs are incorrectly added to egress rules, allowing traffic to reach ports that should be blocked by NetworkPolicy.

When a NetworkPolicy restricts egress to a specific numeric port (e.g., port: 80) using a podSelector, and a matching Service uses a named targetPort (e.g., targetPort: http) backed by pods that resolve that named port to different container ports (e.g., pod A has http=80, pod B has http=8080), the Service ClusterIP gets added to the PolicyEndpoint. Since the eBPF agent enforces policy at the TC hook (pre-DNAT), it sees ClusterIP:80 as allowed, but kube-proxy may DNAT the traffic to pod B:8080, bypassing the policy's port restriction.

This PR skips adding the Service ClusterIP to the PolicyEndpoint when all three conditions are met: the policy uses a numeric port, the service uses a named targetPort, and pods behind the service resolve that named port inconsistently to container ports not covered by the policy. Cases where the policy itself uses a named port are unaffected — that's intentional per-pod resolution by design.

bug fix

- tried this specific Service named port re-mapped case and validated its skipped (UT and manual test)
- ran cyclonus test

**Automation added to e2e**:

**Will this PR introduce any new dependencies?**: NO

**Will this break upgrades or downgrades. Has updating a running cluster been tested?**: NO

Yes, Skipping ServiceIP addition in PolicyEndpoint where its re-mapped to different container ports and if not allowed by Network Policy

```release-note

```

### DIFF
--- a/pkg/resolvers/endpoints.go
+++ b/pkg/resolvers/endpoints.go
@@ -432,18 +432,11 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
 			continue
 		}
 
-		// Fetch pods for this service once - reused for both port bypass issue check and port resolution
-		svcPods, err := r.getPodsForService(ctx, &svc)
-		if err != nil {
-			r.logger.Error(err, "Unable to list pods for service", "service", k8s.NamespacedName(&svc))
-			continue
-		}
-
-		// Check for issue: named service targetPort + inconsistent pod container ports which Network Policy doesn't allow
-		if r.hasNamedPortBypassIssue(&svc, ports, svcPods) {
+		// Check for issue: named service targetPort + pod container ports not allowed by policy
+		if r.hasNamedPortBypassIssue(ctx, &svc, ports) {
 			r.logger.Info("skipping service due to named port bypass issue",
 				"serviceName", svc.Name, "serviceNamespace", svc.Namespace,
-				"reason", "service uses named targetPort with pods resolving to different container ports which policy doesn't allow")
+				"reason", "service uses named targetPort with pods resolving to container ports not allowed by policy")
 			continue
 		}
 
@@ -451,7 +444,7 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
 		for _, port := range ports {
 			var portPtr *int32
 			if port.Port != nil {
-				portVal, err := r.getMatchingServicePort(&svc, port.Port, *port.Protocol, svcPods)
+				portVal, err := r.getMatchingServicePort(ctx, &svc, port.Port, *port.Protocol)
 				if err != nil {
 					r.logger.Info("Unable to lookup service port", "err", err)
 					continue
@@ -495,14 +488,18 @@ func (r *defaultEndpointsResolver) getPodsForService(ctx context.Context, svc *c
 // hasNamedPortBypassIssue checks if adding a Service ClusterIP would create a bypass.
 // The issue occurs when:
 // 1. The Service uses a NAMED targetPort (e.g., targetPort: http)
-// 2. Pods behind the Service resolve that named port to DIFFERENT container ports not allowed by NetworkPolicy ports
-func (r *defaultEndpointsResolver) hasNamedPortBypassIssue(svc *corev1.Service,
-	policyPorts []networking.NetworkPolicyPort, svcPods []corev1.Pod) bool {
-	// Collect numeric ports allowed by the policy
-	allowedPorts := make(map[int32]bool)
+// 2. Any pod behind the Service resolves that named port to a container port not allowed by NetworkPolicy ports
+func (r *defaultEndpointsResolver) hasNamedPortBypassIssue(ctx context.Context, svc *corev1.Service,
+	policyPorts []networking.NetworkPolicyPort) bool {
+	// Collect numeric port+protocol pairs allowed by the policy
+	type portProto struct {
+		port     int32
+		protocol corev1.Protocol
+	}
+	allowedPorts := make(map[portProto]bool)
 	for _, port := range policyPorts {
-		if port.Port != nil && port.Port.Type == intstr.Int {
-			allowedPorts[port.Port.IntVal] = true
+		if port.Port != nil && port.Port.Type == intstr.Int && port.Protocol != nil {
+			allowedPorts[portProto{port: port.Port.IntVal, protocol: *port.Protocol}] = true
 		}
 	}
 	// No numeric policy ports means named ports in policy (intentional per-pod resolution) or all ports allowed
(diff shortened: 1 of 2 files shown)


yash97:
Do you think we should call this out in readme too, in this pr? So we do not have to read code to find this behavior quirk.

viveksb007:
> Do you think we should call this out in readme too, in this pr? So we do not have to read code to find this behavior quirk.

I don't think its necessary as its not common to use in this kind of setup. If we find such queries coming up in future, we can add it to README

pkg/resolvers/endpoints.go
```diff
@@ -432,11 +432,26 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
...
 
+		// Fetch pods for this service once - reused for both port bypass issue check and port resolution
+		svcPods, err := r.getPodsForService(ctx, &svc)
+		if err != nil {
+			r.logger.Error(err, "Unable to list pods for service", "service", k8s.NamespacedName(&svc))
+			continue
+		}
+
+		// Check for issue: named service targetPort + inconsistent pod container ports which Network Policy doesn't allow
+		if r.hasNamedPortBypassIssue(&svc, ports, svcPods) {
```
yash97: instead of inconsistent, it should be named port matches to different ports in backend port.

pkg/resolvers/endpoints.go
```diff
@@ -432,11 +432,26 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
...
+			r.logger.Info("skipping service due to named port bypass issue",
+				"serviceName", svc.Name, "serviceNamespace", svc.Namespace,
```
yash97, line 438: tbf, this says what will be consequence here. It does not say what current state is which making us ignoring this service ip.

pkg/resolvers/endpoints.go
```diff
@@ -432,11 +432,26 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
 			continue
 		}
 
+		// Fetch pods for this service once - reused for both port bypass issue check and port resolution
+		svcPods, err := r.getPodsForService(ctx, &svc)
```
yash97: can we fetch pods, after we confirm it has named ports?

viveksb007: i am assuming from "it has named ports", you mean svc has named port. we were fetching pods in the code post this anyways for existing logic to add svc IP, so i just uplifted that logic to fetch once and use for new check and existing check

pkg/resolvers/endpoints.go
viveksb007: that is what `inconsistent pod container ports` means here

pkg/resolvers/endpoints.go
viveksb007: this is not clear, can you provide details on what you are expecting here?

pkg/resolvers/endpoints.go
yash97: Yeah i was hoping if can avoid get calls if not necessary. its fine.

yash97 (dismissed):

pkg/resolvers/endpoints.go
```diff
@@ -461,33 +476,100 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
...
 		Namespace:     svc.Namespace,
 	}); err != nil {
-		r.logger.Info("Unable to List Pods", "err", err)
-		return 0, err
+		return nil, err
+	}
+	return podList.Items, nil
+}
+
+// hasNamedPortBypassIssue checks if adding a Service ClusterIP would create a bypass.
```
jayanthvn, line 488: Thanks for adding this. This helps future readers..

pkg/resolvers/endpoints.go
jayanthvn: It might be a good optimization which can be done later since this use case is not common.

`getPodsForService` (an API call) happens before `hasNamedPortBypassIssue`, but the bypass check can determine it needs to bail out purely from the service spec and policy ports, without ever looking at pods. You could possibly split the check - 

Check early using only svc + policy ports, before fetching pods
```
if !r.couldHaveNamedPortBypassIssue(&svc, ports) {
    // skip pod fetch entirely for this service
}
svcPods, err := r.getPodsForService(ctx, &svc)
```
Where `couldHaveNamedPortBypassIssue` just checks if the service has named targetPorts AND policy has numeric ports — no pods needed. This avoids the pod list API call for services that clearly won't have the issue.

viveksb007: initially i had separate pod List call for this issue check as we can't be sure of inconsistent port issues unless we also take a look at the pod ports behind it.

we can do early exit based of if named port doesn't on svc. But for cases where named port exists in Svc, we need to check pods behind it.

without this PR changes, the code will go ahead and make list pod call inside fn `getMatchingServicePort`, so I uplifted that call here, so we make it once here. (as this List call will be served from the Informer cache, so its not actual API call).

code pointer (uplifted this part) - https://github.com/aws/amazon-network-policy-controller-k8s/blob/9d8e5d373dd78fd3d332b3a1513320598838c050/pkg/resolvers/endpoints.go#L479

pkg/resolvers/endpoints.go
```diff
@@ -461,33 +476,100 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
...
+					}
+					portResolutions[cp.Name][cp.ContainerPort] = true
+				}
+			}
+		}
+	}
+
+	// Check each named targetPort for inconsistent resolution to disallowed ports
+	for portName, containerPorts := range portResolutions {
+		if len(containerPorts) > 1 {
```
jayanthvn: Might be I'm missing something here, So for `len == 1`, we never checks if that single port is allowed. Say if the policy allows port 8080 and the container port is 80, the ClusterIP should not be added — the traffic would reach port 80 on the pod which the policy doesn't permit. We can just drop this check - `if len(containerPorts) > 1 ` ?

jayanthvn: Synced up offline, `getMatchingServicePort` in the caller would fail to find a matching port and the ClusterIP wouldn't be added anyway.

jayanthvn (dismissed):
lgtm

pkg/resolvers/endpoints.go
```diff
@@ -461,33 +476,100 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
...
+			for _, cp := range container.Ports {
+				if namedTargetPorts[cp.Name] {
+					if portResolutions[cp.Name] == nil {
+						portResolutions[cp.Name] = make(map[int32]bool)
+					}
+					portResolutions[cp.Name][cp.ContainerPort] = true
+				}
+			}
+		}
+	}
```
shiv-amz: My concern is performance degradation due to this O(N^3) loop, which is added just for validation purpose. One option is to refactor such that we return early, when we find a namedPort which is not in allowedPorts.

shiv-amz (approved):

pkg/resolvers/endpoints.go
viveksb007: with current overall implementation, we can't get rid of this complexity without major architectural changes. Small nit on the complexity, instead of `O(N^3)`, we can call it `O(pods * containers per pod * ports per container)`, apart from pods, other 2 will have reasonable upper bounds.

Updated the code to check for any container port if not allowed by NP, it returns early. So this optimizes the happy path of detection. For cases where the issue doesn't exist, it will still see worst case.

pkg/resolvers/endpoints.go
viveksb007: moved common List pod API call to their respective places as in some cases we have signals to not make that List API call even from cache and skip early.

jayanthvn (approved):

shiv-amz (approved):

PR #211, opened 2026-02-17
Increasing log level for logs that are not necessary for debugging on prod

nakshah87:
**What type of PR is this?**
With an increase in pod scale, and high pod churn the "Pod and policy namespace mismatch" log was causing an increase in disk utilization.
- The code logic walks every policy for every pod CRUD operation and prints this during mismatch.

Optimizing the following logs to not print in a loop but consolidate the info in a list and print once at the end of the loop.
* skipping headless service when populating EndpointInfo
* skipping pod selector mismatched service when populating EndpointInfo

**Automation added to e2e**:

**Does this PR introduce any user-facing change?**: No

### DIFF
--- a/pkg/resolvers/endpoints.go
+++ b/pkg/resolvers/endpoints.go
@@ -405,6 +405,11 @@ func (r *defaultEndpointsResolver) createPodLabelSelector(ls *metav1.LabelSelect
 func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Context, ls *metav1.LabelSelector, namespace string,
 	ports []networking.NetworkPolicyPort) []policyinfo.EndpointInfo {
 	var networkPeers []policyinfo.EndpointInfo
+	var headlessServiceList []string
+	var nonMatchingPodSelectorList []string
+	var headlessServiceCount int
+	var nonMatchingPodSelectorCount int
+	const maxSampleSize = 50
 	if ls == nil {
 		ls = &metav1.LabelSelector{}
 	}
@@ -423,12 +428,20 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
 	for _, svc := range svcList.Items {
 		// do not add headless services to policy endpoints
 		if k8s.IsServiceHeadless(&svc) {
-			r.logger.Info("skipping headless service when populating EndpointInfo", "serviceName", svc.Name, "serviceNamespace", svc.Namespace)
+			if headlessServiceCount < maxSampleSize {
+				headlessServiceList = append(headlessServiceList, fmt.Sprintf("%s/%s", svc.Namespace, svc.Name))
+			}
+			headlessServiceCount++
+			r.logger.V(1).Info("skipping headless service when populating EndpointInfo", "serviceName", svc.Name, "serviceNamespace", svc.Namespace)
 			continue
 		}
 		// do not add services if their pod selector is not matching with the pod selector defined in the network policy
 		if !svcSelector.Matches(labels.Set(svc.Spec.Selector)) {
-			r.logger.Info("skipping pod selector mismatched service when populating EndpointInfo", "serviceName", svc.Name, "serviceNamespace", svc.Namespace, "expectedPS", svcSelector)
+			if nonMatchingPodSelectorCount < maxSampleSize {
+				nonMatchingPodSelectorList = append(nonMatchingPodSelectorList, fmt.Sprintf("%s/%s (expectedPS: %s)", svc.Namespace, svc.Name, svcSelector.String()))
+			}
+			nonMatchingPodSelectorCount++
+			r.logger.V(1).Info("skipping pod selector mismatched service when populating EndpointInfo", "serviceName", svc.Name, "serviceNamespace", svc.Namespace, "expectedPS", svcSelector)
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/resolvers/endpoints.go
```diff
@@ -423,12 +423,12 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
 	for _, svc := range svcList.Items {
 		// do not add headless services to policy endpoints
 		if k8s.IsServiceHeadless(&svc) {
-			r.logger.Info("skipping headless service when populating EndpointInfo", "serviceName", svc.Name, "serviceNamespace", svc.Namespace)
+			r.logger.V(1).Info("skipping headless service when populating EndpointInfo", "serviceName", svc.Name, "serviceNamespace", svc.Namespace)
```
oliviassss, line 435: This log is brought to Info() level on purpose  to bring observability due to a previous prod issue. 
Please revert unless it's causing issue in prod currently.
https://github.com/aws/amazon-network-policy-controller-k8s/pull/62

pkg/resolvers/endpoints.go
```diff
@@ -423,12 +423,12 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
...
 			continue
 		}
 		// do not add services if their pod selector is not matching with the pod selector defined in the network policy
 		if !svcSelector.Matches(labels.Set(svc.Spec.Selector)) {
-			r.logger.Info("skipping pod selector mismatched service when populating EndpointInfo", "serviceName", svc.Name, "serviceNamespace", svc.Namespace, "expectedPS", svcSelector)
+			r.logger.V(1).Info("skipping pod selector mismatched service when populating EndpointInfo", "serviceName", svc.Name, "serviceNamespace", svc.Namespace, "expectedPS", svcSelector)
```
oliviassss, line 444: same here

pkg/resolvers/endpoints.go
```diff
@@ -458,6 +458,7 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
 			Ports: portList,
 		})
 	}
+	r.logger.Info("Network peers appended ", networkPeers)
```
oliviassss: can you elaborate why we need to add this line? I think networkPeers itself may not be very useful it just show svc ip and port.

parlakisik (commented):
Could you provide more context in the PR details?
Why do you characterize this as noisy? How much did the log volume change after this improvement? Are there any additional logs being generated?

pkg/resolvers/endpoints.go
nakshah87: I added the above log since I had changed the other 2 logs to V(1) level. Since the above logs are being logged in a for loop, there would be multiple logs for each headless service or pod selecter that is skipped.
With this we can have a consolidated list of networkPeers that were finally picked and have one log for each time this function is called.

oliviassss: I believe the above 2 logs are added intentionally, because we don't want the controller to silently skip headless services or services with mismatched selectors. Simply logging the finalized `networkPeers` at the end wouldn't convey why certain services were skipped.

But I think consolidating the log lines makes sense. can we maintain 2 lists in the for loop, `headlessServices` and `mismatchedServices`, if they're non empty, we add a single log to show which the controller skipped.

nakshah87: Made changes to consolidate the logs.

nakshah87:
> Could you provide more context in the PR details? Why do you characterize this as noisy? How much did the log volume change after this improvement? Are there any additional logs being generated?

@parlakisik I updated the Description and provided more context.

pkg/resolvers/endpoints.go
```diff
@@ -466,90 +462,12 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
 			Ports: portList,
 		})
 	}
+	r.logger.Info("headlessServiceList", headlessServiceList)
```
oliviassss: nit: with consolidating log, we can just remove L429. can add more clear log message here like `r.logger.Info("skipping headless services: ", headlessServiceList)`

pkg/resolvers/endpoints.go
```diff
@@ -466,90 +462,12 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
...
+	r.logger.Info("nonMatchingPodSelectorList", nonMatchingPodSelectorList)
+	r.logger.Info("Network peers appended ", networkPeers)
```
oliviassss: L467 is not needed, since it may add a very large chunk of log lines in every reconcile. we have the skipping list, so it should be fine.

pkg/resolvers/endpoints.go
```diff
@@ -466,90 +462,12 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
 			Ports: portList,
 		})
 	}
+	r.logger.Info("headlessServiceList", headlessServiceList)
+	r.logger.Info("nonMatchingPodSelectorList", nonMatchingPodSelectorList)
```
oliviassss: same here. remove L435 and add more clear log message here.

pkg/resolvers/endpoints.go
nakshah87: Done

pkg/resolvers/endpoints.go
nakshah87: Done

pkg/resolvers/endpoints.go
nakshah87: Done

oliviassss (dismissed):

pkg/resolvers/endpoints.go
```diff
@@ -423,12 +428,20 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
...
 			continue
 		}
 		// do not add services if their pod selector is not matching with the pod selector defined in the network policy
 		if !svcSelector.Matches(labels.Set(svc.Spec.Selector)) {
-			r.logger.Info("skipping pod selector mismatched service when populating EndpointInfo", "serviceName", svc.Name, "serviceNamespace", svc.Namespace, "expectedPS", svcSelector)
+			if nonMatchingPodSelectorCount < maxSampleSize {
+				nonMatchingPodSelectorList = append(nonMatchingPodSelectorList, fmt.Sprintf("%s/%s (expectedPS: %s)", svc.Namespace, svc.Name, svcSelector.String()))
+			}
+			nonMatchingPodSelectorCount++
+			r.logger.V(1).Info("skipping pod selector mismatched service when populating EndpointInfo", "serviceName", svc.Name, "serviceNamespace", svc.Namespace, "expectedPS", svcSelector)
```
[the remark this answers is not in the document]
oliviassss: V(1) log is retained for testing will not affect production. should be fine.

oliviassss (approved):

PR #213, opened 2026-02-20
added change timestamp in PE and CPE annotation

viveksb007:
**What type of PR is this?**
feature

Adds a `networking.k8s.aws/last-change-trigger-time` annotation to PolicyEndpoint and ClusterPolicyEndpoint resources. This annotation is set whenever NPC creates or updates these resources, recording the timestamp in RFC3339Nano format.

This enables the Network Policy Agent (NPA) to compute end-to-end policy programming latency by comparing when the controller made changes versus when they were applied. The pattern mirrors kube-proxy's endpoints.kubernetes.io/last-change-trigger-time annotation on EndpointSlice.

tested by deployment. 

```
kind: PolicyEndpoint
metadata:
  annotations:
    networking.k8s.aws/last-change-trigger-time: "2026-02-19T23:51:44.942777624Z"
  creationTimestamp: "2026-02-19T23:16:34Z"
  generateName: kwok-scale-test-policy-91-
  generation: 30
  name: kwok-scale-test-policy-91-hwj5w
  namespace: scale-test
```

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/policyendpoints/manager.go
+++ b/pkg/policyendpoints/manager.go
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"strconv"
+	"time"
 
 	"golang.org/x/exp/maps"
 
@@ -23,6 +24,30 @@ import (
 	"github.com/aws/amazon-network-policy-controller-k8s/pkg/k8s"
 	"github.com/aws/amazon-network-policy-controller-k8s/pkg/resolvers"
 )
+
+const (
+	// LastChangeTriggerTimeAnnotation is set on PolicyEndpoint/ClusterPolicyEndpoint
+	// to track when NPC last created or updated the resource. NPA reads this to
+	// compute E2E policy programming latency (same pattern as kube-proxy's
+	// endpoints.kubernetes.io/last-change-trigger-time on EndpointSlice).
+	LastChangeTriggerTimeAnnotation = "networking.k8s.aws/last-change-trigger-time"
+)
+
+// setLastChangeTriggerTime sets the E2E latency annotation on a PolicyEndpoint.
+func setLastChangeTriggerTime(pe *policyinfo.PolicyEndpoint) {
+	if pe.Annotations == nil {
+		pe.Annotations = make(map[string]string)
+	}
+	pe.Annotations[LastChangeTriggerTimeAnnotation] = time.Now().UTC().Format(time.RFC3339Nano)
+}
+
+// setLastChangeTriggerTimeOnCPE sets the E2E latency annotation on a ClusterPolicyEndpoint.
+func setLastChangeTriggerTimeOnCPE(cpe *policyinfo.ClusterPolicyEndpoint) {
+	if cpe.Annotations == nil {
+		cpe.Annotations = make(map[string]string)
+	}
+	cpe.Annotations[LastChangeTriggerTimeAnnotation] = time.Now().UTC().Format(time.RFC3339Nano)
+}
 
 // PolicyMetadata contains the minimal metadata needed for creating PolicyEndpoints
 type PolicyMetadata struct {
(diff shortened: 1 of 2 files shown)


pkg/policyendpoints/manager.go
```diff
@@ -163,26 +190,28 @@ func (m *policyEndpointsManager) ReconcileANP(ctx context.Context, anp *policyin
...
-	for _, policyEndpoint := range updateList {
+	for i := range updateList {
 		oldRes := &policyinfo.PolicyEndpoint{}
-		if err := m.k8sClient.Get(ctx, k8s.NamespacedName(&policyEndpoint), oldRes); err != nil {
+		if err := m.k8sClient.Get(ctx, k8s.NamespacedName(&updateList[i]), oldRes); err != nil {
 			return err
 		}
-		if equality.Semantic.DeepEqual(oldRes.Spec, policyEndpoint.Spec) {
-			m.logger.V(1).Info("ANP policy endpoint already up to date", "id", k8s.NamespacedName(&policyEndpoint))
+		if equality.Semantic.DeepEqual(oldRes.Spec, updateList[i].Spec) {
```
Pavani-Panakanti, line 206: why don't we have this deep equal check for ReconcileCNP ? Am I missing anything ?
In case of CNP will annotation be added though generated spec is same as earlier ?

viveksb007: probably looks like a miss, based on semantic check, i believe we should be able to skip the CPE update similar to PE update.

annotations are not part of the spec.

Pavani-Panakanti: For CPE --> we are missing deep equal check. So we will be adding annotation on CPE irrespective of there is change in CPE or not. I was wondering if there will be any additional api server or agent watch calls due to this but I see we are doing that irrespective of this new annotation. 
Not related to this PR but probably we should add that deep equal check to CPE too

Pavani-Panakanti:
This is still not end to end metric right. We will be missing any time between when cx applied the NP to NPC got the reconcile right ? -> api server to NPC watch delivery and NPC queue wait time. Do we already have these or any plans to add them

Pavani-Panakanti:
Overall PR looks good to me. Will wait for co pilot comments to be addressed

pkg/policyendpoints/manager.go
```diff
@@ -24,6 +25,30 @@ import (
...
+	}
+	pe.Annotations[LastChangeTriggerTimeAnnotation] = time.Now().UTC().Format(time.RFC3339Nano)
+}
+
+// setLastChangeTriggerTimeOnCPE sets the E2E latency annotation on a ClusterPolicyEndpoint.
+func setLastChangeTriggerTimeOnCPE(cpe *policyinfo.ClusterPolicyEndpoint) {
+	if cpe.Annotations == nil {
+		cpe.Annotations = make(map[string]string)
+	}
+	cpe.Annotations[LastChangeTriggerTimeAnnotation] = time.Now().UTC().Format(time.RFC3339Nano)
```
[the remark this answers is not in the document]
viveksb007: i had checked this before and decided keep similar functions because of different types of PE and CPE

viveksb007:
> This is still not end to end metric right. We will be missing any time between when cx applied the NP to NPC got the reconcile right ? -> api server to NPC watch delivery and NPC queue wait time. Do we already have these or any plans to add them

its not true E2E, rather covering latencies from NPC to NPA. For NPC, we can look at its WQ and controller-runtime metrics to get additional signals. NP multi dimension makes it tricky to calculate true E2E but this is a good starting point to measure Cx experience latencies, we can explore further if we can inject timestamps coming from the trigger like Svc, Pods, Ns, etc to measure true E2E.

pkg/policyendpoints/manager.go
viveksb007: added in new commit in this PR

pkg/policyendpoints/manager_test.go
```diff
@@ -1107,6 +1108,49 @@ func (m *mockCNPEndpointsResolver) ResolveClusterNetworkPolicy(ctx context.Conte
 	return m.ingressRules, m.egressRules, m.podEndpoints, nil
 }
 
+func Test_setLastChangeTriggerTime(t *testing.T) {
+	t.Run("sets annotation on PE with nil annotations", func(t *testing.T) {
+		pe := &policyinfo.PolicyEndpoint{}
+		before := time.Now().UTC()
+		setLastChangeTriggerTime(pe)
+		after := time.Now().UTC()
```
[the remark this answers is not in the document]
viveksb007: no existing test for overall Reconcile

pkg/policyendpoints/manager.go
```diff
@@ -699,18 +728,28 @@ func (m *policyEndpointsManager) ReconcileCNP(ctx context.Context, cnp *policyin
...
+		oldRes := &policyinfo.ClusterPolicyEndpoint{}
+		if err := m.k8sClient.Get(ctx, k8s.NamespacedName(&updateList[i]), oldRes); err != nil {
+			return err
+		}
+		if equality.Semantic.DeepEqual(oldRes.Spec, updateList[i].Spec) {
+			m.logger.V(1).Info("Cluster policy endpoint already up to date", "id", k8s.NamespacedName(&updateList[i]))
+			continue
+		}
+		setLastChangeTriggerTimeOnCPE(&updateList[i])
+		if err := m.k8sClient.Update(ctx, &updateList[i]); err != nil {
```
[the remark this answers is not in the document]
viveksb007: unrelated to this PR. Its doing pattern matching with above PE logic.

Pavani-Panakanti (approved):

jaydeokar (approved):
lgtm

PR #219, opened 2026-03-09
Perf improvement and endpoints refactoring

viveksb007:
**What type of PR is this?**
improvements

1. Unnecessary PolicyEndpoint updates due to order-sensitive equality checks
    - The controller was using equality.Semantic.DeepEqual to compare PolicyEndpoint specs, which is order-sensitive for slices. This caused spurious updates to PolicyEndpoint resources when
Ingress/Egress/PodSelectorEndpoints were semantically identical but in different order.
    - Adds a new pkg/equality package with EqualPolicyEndpointSpec and EqualClusterPolicyEndpointSpec that perform order-insensitive comparison of Ingress, Egress, PodSelectorEndpoints,
Ports, Except, and PodIsolation slices.

2. Excessive Informer cache calls from per-service pod list lookups
    - getMatchingServiceClusterIPs was calling getPodsForService (which does a client.List which is server from Informer cache but results is copying over all Pods in that namespace) separately for each service, both in hasNamedPortBypassIssue and
getMatchingServicePort. For N services, this could result in up to 2N pod list calls. This frequent copying from informer cache is results in a lot GC (During profiling almost 40% time goes in GC for cases where there are 3000 svc and 2000 Pods)
    - Refactored to use a lazy getAllPods closure that fetches all pods in the namespace once and caches the result. Pod filtering by service selector is now done in-memory. This reduces Informer cache call from O(N) to O(1) pod list calls.

3. Minor cleanups
    - Replaced github.com/pkg/errors with fmt.Errorf in pod_utils.go and service_utils.go (error.Errorf print entire stack trace which contributes to unnecessary logging)
    - Upgraded Go version from 1.25.7 to 1.26 (to use go version improvements and greentea gc https://go.dev/doc/go1.26)
    - Renamed function parameters for clarity (port → npPort, protocol → npProtocol, ls → npPodSelector)
    - Added TODO comment about hash-based bin packing potentially generating different hashes for semantically equal EndpointInfo (will check as follow up)

This PR fixes the excessive GC for large pod and svc objects. And also reduces the duration of Reconcile. During test, i was able to bring down latency from 14s to 2s for certain workload (which had 2000 Pods and 3000 Svc and 1 Network Policy which was selecting all Pods. Note: it was also selecting all Svc IPs because NPC adds svc IPs as allowlist [code pointer](https://github.com/aws/amazon-network-policy-controller-k8s/blob/ab8c82eaf17333b9cfa4704475c098acb8322a25/pkg/resolvers/endpoints.go#L93)).

- UT and cyclonus tests

**Automation added to e2e**:

```release-note

```


pkg/resolvers/endpoints.go
```diff
@@ -425,9 +425,27 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
 		r.logger.Info("Unable to list services", "err", err)
 		return nil
 	}
-	for _, svc := range svcList.Items {
+
+	// Lazily fetch all pods in namespace once to avoid repeated cache iterations
+	var allPods []corev1.Pod
+	getAllPods := func() ([]corev1.Pod, error) {
+		if allPods != nil {
+			return allPods, nil
```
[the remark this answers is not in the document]
viveksb007: this is fine, if there are no pods, cache lookup will not cost anything. this is mainly done to avoid copying objects from informer cache again and again in a loop

pkg/resolvers/endpoints.go
```diff
@@ -425,9 +425,27 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
 		r.logger.Info("Unable to list services", "err", err)
 		return nil
 	}
-	for _, svc := range svcList.Items {
+
+	// Lazily fetch all pods in namespace once to avoid repeated cache iterations
+	var allPods []corev1.Pod
+	getAllPods := func() ([]corev1.Pod, error) {
```
oliviassss, line 431: I'm a bit concerned on the tradeoff for this lazily fetch all pods in namespace. it reduced the API call load by loading once for all pods in the namespace into memory, rather than
filtering by service selector. this makes sense when the pod and matched svc scale are quite  tie in ns (say 1k pods, 500svc).
However, any concern with large namespace that has many pods but few matching services (say 5k pods vs only 10 match svc).  now the controller will load 5k pods at once and filter them in a loop. any concern for the memory and cpu?

pkg/equality/semantic.go
```diff
@@ -0,0 +1,225 @@
...
+		return false
+	}
+	if !equality.Semantic.DeepEqual(a.PolicyRef, b.PolicyRef) {
+		return false
+	}
+	if !equalPodIsolation(a.PodIsolation, b.PodIsolation) {
+		return false
+	}
+
+	// Unordered slice fields.
```
oliviassss, line 29: nice catch on the unorder slice fileds that causing unnecessary reconciliations!

I'm wondering if we should maintain consistent ordering of `Ingress/Egress/PodSelectorEndpoints` slices instead of introducing custom semantic equality logic.

My concern is the long-term maintenance burden: every time we add new fields to the CRDs, we need to remember to update `pkg/equality/semantic.go`. We're just one unaware engineer away from missing this and reintroducing the spurious update problem.

pkg/resolvers/endpoints.go
viveksb007: this looks API call but is served via Informer cache.

we are lazily fetching the pods meaning the call to informer cache will happen only when it needs to iterate on the pods (in `hasNamedPortBypassIssue` or `getMatchingServicePort`, will fetch once if pod iteration is included in both).

these pods are maintained in the informer cache and this List call results in copying those objects from the informer cache to the caller. so we are doing that once and then filtering in-memory using selector matching. 

the concern is valid, in cases where 5K pods and only 10 pods match svc, we would have copied all 5K pods in-mem and then filtered them out instead of informer cache filtering them. From CPU perspective, both are almost same as iteration and matching will still be happening under the hood in controller-runtime code.

I haven't seen any issues during testing where we were keeping large number of pods (as these are not complete pods rather some metadata of the pod, ref - https://github.com/aws/amazon-network-policy-controller-k8s/blob/0be0c0e5d43347fb7a7c1d0e6b908ecddfd59f2a/pkg/config/runtime_config.go#L96 )

these pods are scoped to getMatchingServiceClusterIPs, GC'd on return.

pkg/equality/semantic.go
viveksb007: it seems long term maintenance issue is in both cases (sort at source of object generation or deep equal check), if there is a new field added and not considered for equality comparison while generating object, we will end up in same situation. so whoever is working on this needs to be aware about consistent generation of objects.

generating consistent `Ingress/Egress/PodSelectorEndpoints` means doing sorting at many places deep in the code where this generation is happening. Or do a deep compare outside, this seems better from readability and time complexity perspective (additionally consistent ordering will happen for all objects whereas this deep check will happen for updates).

viveksb007: ~~TODO - synced offline, add a UT to check deep compare fields with object using reflection?~~ 
Done

New field addition without DeepCheck will fail UT
```
=== RUN   TestPortKey_CoversAllFields
    /workplace/viveksbh/NP-Perf/amazon-network-policy-controller-k8s/pkg/equality/semantic_test.go:405:
                Error Trace:    /workplace/viveksbh/NP-Perf/amazon-network-policy-controller-k8s/pkg/equality/semantic_test.go:405
                Error:          elements differ

                                extra elements in list B:
                                ([]interface {}) (len=1) {
                                 (string) (len=8) "NewThing"
                                }


                                listA:
                                ([]string) (len=3) {
                                 (string) (len=8) "Protocol",
                                 (string) (len=4) "Port",
                                 (string) (len=7) "EndPort"
                                }


                                listB:
                                ([]string) (len=4) {
                                 (string) (len=8) "Protocol",
                                 (string) (len=4) "Port",
                                 (string) (len=7) "EndPort",
                                 (string) (len=8) "NewThing"
                                }
                Test:           TestPortKey_CoversAllFields
                Messages:       Port fields changed — update portKey in semantic.go
--- FAIL: TestPortKey_CoversAllFields (0.00s)
```

oliviassss (approved):

PR #225, opened 2026-03-19
fixed: bypass logic for same named port containers in a pod

viveksb007:
**What type of PR is this?**
bugfix

follow up from https://github.com/aws/amazon-network-policy-controller-k8s/pull/223

when containers in a pod have same named port resolving to different numeric ports. Svc is resolved to first container matching as its Endpoint but NPC bypass logic was checking all containers named ports and resulting in un-necessary skip

- UT and manual test

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/resolvers/endpoints.go
+++ b/pkg/resolvers/endpoints.go
@@ -550,26 +550,19 @@ func (r *defaultEndpointsResolver) hasTargetPortBypass(svc *corev1.Service, matc
 		if !svcSelector.Matches(labels.Set(pod.Labels)) {
 			continue
 		}
-		for j := range pod.Spec.Containers {
-			container := &pod.Spec.Containers[j]
-			for k := range container.Ports {
-				cp := container.Ports[k]
-				if cp.Name == targetPortName {
-					cpProto := cp.Protocol
-					if cpProto == "" {
-						cpProto = corev1.ProtocolTCP
-					}
-					if !isNumericPortAllowed(cp.ContainerPort, cpProto, policyPorts) {
-						r.logger.Info("Named port resolves to disallowed container port",
-							"service", k8s.NamespacedName(svc),
-							"pod", k8s.NamespacedName(pod),
-							"container", container.Name,
-							"portName", cp.Name,
-							"containerPort", cp.ContainerPort, "protocol", cpProto)
-						return true
-					}
-				}
-			}
+		// Resolve the named port using the first matching container port, matching
+		// Kubernetes EndpointSlice controller behavior for svc resolution.
+		cp, found := firstNamedPort(pod, targetPortName, protocol)
+		if !found {
+			continue
+		}
+		if !isNumericPortAllowed(cp.ContainerPort, protocol, policyPorts) {
+			r.logger.Info("Named port resolves to disallowed container port",
+				"service", k8s.NamespacedName(svc),
+				"pod", k8s.NamespacedName(pod),
+				"portName", cp.Name,
+				"containerPort", cp.ContainerPort, "protocol", protocol)
+			return true
 		}
 	}
 	return false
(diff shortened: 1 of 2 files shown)


pkg/resolvers/endpoints.go
```diff
@@ -609,6 +602,25 @@ func isNumericPortAllowed(port int32, protocol corev1.Protocol, policyPorts []ne
 	return false
 }
 
+// firstNamedPort returns the first container port matching the given name and protocol in a pod.
+// This mirrors how Kubernetes EndpointSlice controller resolves named targetPorts — it picks the
+// first matching container port across all containers in declaration order.
+func firstNamedPort(pod *corev1.Pod, portName string, protocol corev1.Protocol) (corev1.ContainerPort, bool) {
+	for i := range pod.Spec.Containers {
```
parlakisik, line 610: will you bypass  check ephemeral containers .  they may be handled here too (pod.Spec.EphemeralContainers)

pkg/resolvers/endpoints.go
```diff
@@ -609,6 +602,25 @@ func isNumericPortAllowed(port int32, protocol corev1.Protocol, policyPorts []ne
...
+		for j := range pod.Spec.Containers[i].Ports {
```
parlakisik, line 611: instead of using i j variables can we write like 
```
   for _, container := range pod.Spec.Containers {    
       for _, cp := range container.Ports { 
              cpProto := cp.Protocol    
     ....
 ```

Comment (viveksb007) [reply pkg/resolvers/endpoints.go:611]: it was initially this that only but it results in copying the object, they would be copy of `Containers` and `Ports`. As part of performance optimizations, i moved them to index based to avoid copying object often as it was increasing GC times. This code path can be called many times depending on the scale

File: pkg/resolvers/endpoints.go [hunk 3]
```diff
@@ -550,26 +550,19 @@ func (r *defaultEndpointsResolver) hasTargetPortBypass(svc *corev1.Service, matc
...
+		// Resolve the named port using the first matching container port, matching
+		// Kubernetes EndpointSlice controller behavior for svc resolution.
+		cp, found := firstNamedPort(pod, targetPortName, protocol)
+		if !found {
+			continue
+		}
+		if !isNumericPortAllowed(cp.ContainerPort, protocol, policyPorts) {
+			r.logger.Info("Named port resolves to disallowed container port",
+				"service", k8s.NamespacedName(svc),
+				"pod", k8s.NamespacedName(pod),
```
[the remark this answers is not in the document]
viveksb007: no needed as svc chooses first container with matching namedPort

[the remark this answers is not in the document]
Pavani-Panakanti: container name is still helpful here..if not we need to go and check which is the first container in pod spec that matches the port ? Adding container would make debugging easier

pkg/resolvers/endpoints.go
viveksb007: other containers types like `ephemeral`, `init` and `sidecar` won't be included behind svc IP for this bypass issue.

i found a few interesting things while validating this 
- ephemeralContainer can't set ports
```
spec.ephemeralContainers[0].ports: Forbidden: cannot be set for an Ephemeral Container
```
- initContainer needs to exit successfully for main container to start (if main container never starts, pod doesn't become READY and is never added to Service Endpoints)
- sidecar container can have port collision but EndpointSlice controller will choose first matching container and add that behind the svc

Pavani-Panakanti: WDYT about adding a 2 line comment about this and these are intentionally skipped for later ?

parlakisik (dismissed):
lgtm

pkg/resolvers/endpoints.go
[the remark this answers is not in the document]
viveksb007: done

pkg/resolvers/endpoints.go
viveksb007: i think that is too much info to put as a code comment there for something not doing in code. its good to know though

parlakisik (approved):
lgtm

PR #227, opened 2026-04-24
endpoint resolve optimization and refactor

viveksb007:
**What type of PR is this?**
improvement/refactor

NPC reconciliation work in large scale clusters

Improvement items

1. Port Resolution Optimization for Ingress: Ingress port resolution happened per-peer. For each peer in a rule, the resolver would list the destination pods to translate named ports to numbers — even if every port was already numeric and no translation was needed. For IPBlock and allow-all peers, it was worse: each named port triggered its own separate pod listing call.
    1. Fix: Port resolution happens once per rule, before iterating over peers. The resolved port list is then shared across all peers in that rule
2. Parallel Resolution: Ingress, egress, and pod selector endpoints are now resolved concurrently instead of one after another, reducing latency.
3. Refactoring to clearer separation of ingress vs egress logic: Previously, a single generic function handled both directions using a policyType flag with internal branching. Now each direction has its own dedicated path, making the port resolution strategy for each direction explicit (ingress resolves ports from destination pods; egress resolves per source pod).

- cyclonus
- UT

**Automation added to e2e**:

```release-note

```


pkg/resolvers/endpoints.go
```diff
@@ -42,65 +44,252 @@ type defaultEndpointsResolver struct {
...
+			if portInfo != nil {
+				portList = append(portList, *portInfo)
+			}
+		}
+		return portList
+	}
+
+	// Has named ports — resolve from destination pods
+	return r.getIngressRulesPorts(ctx, policy.Namespace, &policy.Spec.PodSelector, ports)
+}
```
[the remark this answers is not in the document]
viveksb007: this is fine, we do semantic deep check with old/new PE (https://github.com/aws/amazon-network-policy-controller-k8s/blob/main/pkg/equality/semantic.go)

pkg/resolvers/endpoints.go
```diff
@@ -42,65 +44,252 @@ type defaultEndpointsResolver struct {
...
+			namespaces, err := r.resolvePeerNamespaces(ctx, policy.Namespace, peer)
+			if err != nil {
+				return nil, fmt.Errorf("resolving egress peers: %w", err)
+			}
+
+			for _, ns := range namespaces {
+				// Pod endpoints: list matching pods, resolve ports per destination pod
+				pods, err := r.listMatchingPods(ctx, ns, peer.PodSelector)
+				if err != nil {
+					r.logger.Info("Unable to List Pods", "err", err)
```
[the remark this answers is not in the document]
viveksb007: existing code, this list call is from cache

pkg/resolvers/endpoints.go
```diff
@@ -42,65 +44,252 @@ type defaultEndpointsResolver struct {
...
+
+			namespaces, err := r.resolvePeerNamespaces(ctx, policy.Namespace, peer)
+			if err != nil {
+				return nil, fmt.Errorf("resolving ingress peers: %w", err)
+			}
+
+			for _, ns := range namespaces {
+				pods, err := r.listMatchingPods(ctx, ns, peer.PodSelector)
+				if err != nil {
+					r.logger.Info("Unable to List Pods", "err", err)
```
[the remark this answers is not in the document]
viveksb007: existing code, this list call is from cache

pkg/resolvers/endpoints.go
```diff
@@ -42,65 +44,252 @@ type defaultEndpointsResolver struct {
...
+		// Step 1: Resolve ports once per rule.
+		// Named ports are resolved against the policy's destination pods.
+		// Numeric and nil (protocol-only) ports are resolved without pod lookups.
+		resolvedPorts := r.resolveIngressPorts(ctx, policy, rule.Ports)
+		if len(rule.Ports) != 0 && len(resolvedPorts) == 0 {
+			r.logger.Info("Couldn't resolve any ports for ingress rule, skipping",
+				"policy", k8s.NamespacedName(policy), "ports", rule.Ports)
+			continue
+		}
+
```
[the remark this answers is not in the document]
viveksb007: will defer this for future if needed. the policy resulting in this is valid YAML but wouldn't be intentional

```
  kind: NetworkPolicy
  metadata:
    name: deny-all-ingress
    namespace: default
  spec:
    podSelector:
      matchLabels:
        app: web
    policyTypes:
      - Ingress
    ingress:
      - from: []          # empty list — no sources allowed
        ports:
          - protocol: TCP
            port: http    # named port — would trigger a pod listing
```

pkg/resolvers/endpoints_test.go
```diff
@@ -1369,47 +1316,41 @@ func TestEndpointsResolver_IncludesHostNetworkPodsInIngressEgressRules(t *testin
...
+					typedList.Items = []corev1.Pod{hostNetworkMonitoringPod, regularMonitoringPod}
+				case "role=database":
+					typedList.Items = []corev1.Pod{hostNetworkDatabasePod, regularDatabasePod}
+				case "role=backend":
+					// Used by both getIngressRulesPorts and computePodSelectorEndpoints.
+					typedList.Items = []corev1.Pod{targetPod}
+				default:
+					t.Errorf("unexpected pod list selector: %q", sel)
+					return fmt.Errorf("unexpected pod list selector: %q", sel)
+				}
```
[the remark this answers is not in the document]
viveksb007: fixed

pkg/resolvers/endpoints.go
```diff
@@ -42,65 +44,252 @@ type defaultEndpointsResolver struct {
...
+	}
+}
+
+// buildIPBlockEndpoint converts an IPBlock peer into an EndpointInfo entry with the given ports.
+func (r *defaultEndpointsResolver) buildIPBlockEndpoint(ipBlock *networking.IPBlock, ports []policyinfo.Port) []policyinfo.EndpointInfo {
+	var except []policyinfo.NetworkAddress
+	for _, ea := range ipBlock.Except {
+		except = append(except, policyinfo.NetworkAddress(ea))
+	}
+	return []policyinfo.EndpointInfo{
```
Pavani-Panakanti: Return is always one EndpointInfo. Should we change return from List to single EndpointInfo struct

pkg/resolvers/endpoints.go
```diff
@@ -42,65 +44,252 @@ type defaultEndpointsResolver struct {
...
 		}
-		resolvedPeers, err := r.resolveNetworkPeers(ctx, policy, rule.To, rule.Ports, networking.PolicyTypeEgress)
-		if err != nil {
-			return nil, errors.Wrap(err, "unable to resolve egress network peers")
-		}
-		resolvedClusterIPs, err := r.resolveServiceClusterIPs(ctx, rule.To, policy.Namespace, rule.Ports)
-		if err != nil {
-			return nil, errors.Wrap(err, "unable to resolve service cluster IPs for egress")
+
+		for _, peer := range rule.To {
```
Pavani-Panakanti, line 218: We already have buildIPBlockEndpoint for ingress, the egress IPBlock path is also doing the same thing. How about precomputing the CIDR port list once before the peer loop and reusing buildIPBlockEndpoint here too ? It will remove the duplicated except/CIDR logic and keeps both directions consistent.

pkg/resolvers/endpoints_test.go
```diff
@@ -935,171 +897,156 @@ func TestEndpointsResolver_ResolveNetworkPeers_NamedIngressPortsIPBlocks(t *test
...
-		Spec: networking.NetworkPolicySpec{
-			PodSelector: metav1.LabelSelector{},
-			PolicyTypes: []networking.PolicyType{networking.PolicyTypeIngress},
-			Ingress: []networking.NetworkPolicyIngressRule{
-				{
-					From: []networking.NetworkPolicyPeer{
-						{
-							IPBlock: &networking.IPBlock{
-								CIDR: "100.64.0.0/16",
+	t.Run("IPBlock with named ports", func(t *testing.T) {
```
Pavani-Panakanti, line 900: Can we add mixed ports as well ? Due to code changes around this it took some time for me to validate if mixed ports behavior is same before and after. As this is most common case I think we should have a test case for it

pkg/resolvers/endpoints.go
```diff
@@ -42,65 +44,252 @@ type defaultEndpointsResolver struct {
...
+		defer wg.Done()
+		egressEndpoints, egressErr = r.computeEgressEndpoints(ctx, policy)
+	}()
+	go func() {
+		defer wg.Done()
+		podSelectorEndpoints, podSelectorErr = r.computePodSelectorEndpoints(ctx, policy)
+	}()
+	wg.Wait()
+
+	if err := stderrors.Join(ingressErr, egressErr, podSelectorErr); err != nil {
```
Pavani-Panakanti, line 72: so even if one errors out we complete other go routines. I don't think these are time taking calls and all are cached list calls so should be good but want to confirm once

pkg/resolvers/endpoints.go
```diff
@@ -42,65 +44,252 @@ type defaultEndpointsResolver struct {
...
+// to all peers. This function resolves ports once per rule, then resolves peers using those ports.
 func (r *defaultEndpointsResolver) computeIngressEndpoints(ctx context.Context, policy *networking.NetworkPolicy) ([]policyinfo.EndpointInfo, error) {
 	var ingressEndpoints []policyinfo.EndpointInfo
 	for _, rule := range policy.Spec.Ingress {
 		r.logger.V(1).Info("computing ingress addresses", "peers", rule.From)
+
+		// Step 1: Resolve ports once per rule.
+		// Named ports are resolved against the policy's destination pods.
+		// Numeric and nil (protocol-only) ports are resolved without pod lookups.
+		resolvedPorts := r.resolveIngressPorts(ctx, policy, rule.Ports)
```
Pavani-Panakanti, line 93: how about adding this comment to be safe for any future changes 
```
 // resolvedPorts is shared across all EndpointInfo for this rule.
  // Downstream consumers MUST NOT mutate the Ports slice.
```

Pavani-Panakanti:
Changes looks good overall

pkg/resolvers/endpoints.go
viveksb007: yes, we wait for all 3 to return. yeah, all are doing work off informer cache only, depending on the scale of ingress/egress/target, they would have associate time cost, but in case of err, this delay doesn't matter much as we won't be updating anything.

in case of no err, we need to wait for all to complete.

pkg/resolvers/endpoints.go
viveksb007: not clear what do you mean by downstream consumers? you mean other functions which are receiving this resolved port slice?

pkg/resolvers/endpoints.go
viveksb007: done, thanks

pkg/resolvers/endpoints.go
viveksb007: updated

pkg/resolvers/endpoints_test.go
viveksb007: added

pkg/resolvers/endpoints.go
Pavani-Panakanti: Yes as we are passing this ref to multiple funtions, if any of them modifies it would have impact on others. I don't see a usecase why we would do that but just for any future extensions or features we might add

viveksb007: these are resolved ports which are our port objects, so i don't see any reason for anyone to add mutation logic there. if even something is added, tests should capture it as it would fail test verification

Pavani-Panakanti: sounds good. Thanks

Pavani-Panakanti (approved):

PR #190, opened 2025-09-08
Fix issue where PolicyEndpoints are not updated correctly when Pods a…

Subzidion:
Fix issue where PolicyEndpoints are not updated correctly when Pods are in a terminal phase (Succeeded or Failed).

**What type of PR is this?**
Bug Fix.

**Which issue does this PR fix**: IP Addresses can be re-assigned after a Pod has reached a terminal phase (Succeeded or Failed) but the Policy Endpoint will not be updated correctly until the Pod is deleted.

**What does this PR do / Why do we need it**: Enforcement is not working correctly for Pods in the Terminal Phase since their IP Addresses have been re-assigned but the Controller has not processed their completion.

**If an issue # is not available please add steps to reproduce and the controller logs**:

**Testing done on this change**:
Added new tests. Confirmed test failed before making change.

BEFORE:

```
?   	github.com/aws/amazon-network-policy-controller-k8s/api/v1alpha1	[no test files]
?   	github.com/aws/amazon-network-policy-controller-k8s/cmd	[no test files]
ok  	github.com/aws/amazon-network-policy-controller-k8s/internal/controllers	(cached)
?   	github.com/aws/amazon-network-policy-controller-k8s/internal/eventhandlers	[no test files]
?   	github.com/aws/amazon-network-policy-controller-k8s/mocks/controller-runtime/client	[no test files]
ok  	github.com/aws/amazon-network-policy-controller-k8s/pkg/config	(cached)
ok  	github.com/aws/amazon-network-policy-controller-k8s/pkg/crd	(cached)
ok  	github.com/aws/amazon-network-policy-controller-k8s/pkg/k8s	0.591s
ok  	github.com/aws/amazon-network-policy-controller-k8s/pkg/policyendpoints	(cached)
--- FAIL: TestEndpointsResolver_ExcludesTerminalPods (0.00s)
    endpoints_test.go:1175:
        	Error Trace:	/Users/carlhilt/work/eks/networking/network-policy/amazon-network-policy-controller-k8s/pkg/resolvers/endpoints_test.go:1175
        	Error:      	"[{ 10.0.0.1 running-pod test-ns} { 10.0.0.2 succeeded-pod test-ns} { 10.0.0.3 failed-pod test-ns}]" should have 1 item(s), but has 3
        	Test:       	TestEndpointsResolver_ExcludesTerminalPods
        	Messages:   	Should only include running pod in PolicyEndpoints
FAIL
[...]

PR #233, opened 2026-06-02
NPC memory improvements

viveksb007:
**What type of PR is this?**

Improvements

NPC controller OOMKills at startup on clusters with many Pending/Terminating/evicted pods. pprof on a repro showed ~96% of live heap is `ObjectMeta.Unmarshal` (annotation/label strings) cached for every pod the controller never acts on.

Two stackable Pod-cache reductions:

  1. **Server-side field selector** on the Pod informer (`pkg/config/runtime_config.go`):
     `status.phase!=Succeeded,status.phase!=Failed,spec.nodeName!=""`. Apiserver filters terminal-phase and unscheduled pods before serializing — they never reach the cache. Phase transitions emit `Deleted`/`Added` watch events, which existing handlers already process correctly.

  2. **Annotation prune in `StripDownPodTransformFunc`** (`pkg/k8s/pod_utils.go`): only annotation NPC reads is `vpc.amazonaws.com/pod-ips`. Cached pods now keep just that one key; labels untouched.

Succeeded/Failed pods are not useful for NPC PE Endpoints https://github.com/aws/amazon-network-policy-controller-k8s/pull/190

- New unit tests for transform annotation pruning; existing tests pass.
- EKS repro (15K bloated Pending pods, 1 GiB limit): baseline OOMKills in 36 s; with fix, controller stable >5 min, 0 restarts.

**Automation added to e2e**:

```release-note

```


pkg/config/runtime_config.go
```diff
@@ -88,13 +89,38 @@ func BuildRestConfig(rtCfg RuntimeConfig) (*rest.Config, error) {
...
+// When a pod transitions out of the matching set (e.g. Running -> Succeeded),
+// the apiserver synthesizes a watch.Deleted event so the controller's existing
+// Delete handler can remove the pod from any PolicyEndpoints. When a pod
+// transitions into the matching set (scheduler assigns nodeName), the apiserver
+// synthesizes a watch.Added event.
+func podCacheFieldSelector() fields.Selector {
+	return fields.AndSelectors(
+		fields.OneTermNotEqualSelector("status.phase", string(corev1.PodSucceeded)),
+		fields.OneTermNotEqualSelector("status.phase", string(corev1.PodFailed)),
+		fields.OneTermNotEqualSelector("spec.nodeName", ""),
```
parlakisik, line 112: the scheduler assigns nodeName quickly in healthy clusters. Could we simplify by dropping this selector?

Did you see high number of the pods without nodenames in your tests ?

viveksb007: good question, this is not targeting to solve healthy cluster but rather problematic ones where there are 2-3 Nodes and 10-20K Pods (most of Pods will be Unscheduled as Nodes would have reached maxPods limit). From NPC perspective, Unscheduled Pods are not useful, thus not needed for Informer cache.

parlakisik (approved):
LGTM

PR #237, opened 2026-06-08
fix: nil pointer dereference in convertToPolicyInfoPortForCIDRs and bump up docker version in gh workflow

oliviassss:
**What type of PR is this?**

bug
1. convertToPolicyInfoPortForCIDRs at endpoints.go:381 unconditionally dereferences port.Protocol (protocol := *port.Protocol). The K8s NetworkPolicy API allows Protocol to be nil (defaults to TCP per spec), but NPC panics when it encounters this. This causes a SIGSEGV crash-loop whenever an ANP/NP has an egress ipBlock CIDR rule with a port that omits the protocol field.

The fix adds a nil guard and defaults to corev1.ProtocolTCP, matching the existing pattern already used at line 521 of the same file.

2. bump up docker version to v3 for gh action failure fix.

Unit test added: TestEndpointsResolver_Resolve/egress_ipBlock_CIDR_with_nil_Protocol_defaults_to_TCP

  Verified the test reproduces the exact same panic against the unfixed code:
  panic: runtime error: invalid memory address or nil pointer dereference
  [signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x1434641]
  resolvers.(*defaultEndpointsResolver).convertToPolicyInfoPortForCIDRs  endpoints.go:381

  And passes with the fix applied:
  --- PASS: TestEndpointsResolver_Resolve/egress_ipBlock_CIDR_with_nil_Protocol_defaults_to_TCP (0.00s)
  PASS

  Full unit test suite passes: ok github.com/aws/amazon-network-policy-controller-k8s/pkg/resolvers 0.029s
**Automation added to e2e**:

```release-note

```

### DIFF
--- a/.github/workflows/pr-tests.yaml
+++ b/.github/workflows/pr-tests.yaml
@@ -55,9 +55,9 @@ jobs:
           check-latest: true
           cache-dependency-path: "**/go.sum"
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v2
+        uses: docker/setup-qemu-action@v3
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
+        uses: docker/setup-buildx-action@v3
       - name: Build Network Policy Controller images
         run: make docker-buildx
   deprecated-apigroups:
--- a/pkg/resolvers/endpoints_test.go
+++ b/pkg/resolvers/endpoints_test.go
@@ -496,6 +496,45 @@ func TestEndpointsResolver_Resolve(t *testing.T) {
 			},
 		},
 		{
+			name: "egress ipBlock CIDR with nil Protocol defaults to TCP",
+			args: args{
+				netpol: &networking.NetworkPolicy{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "nil-proto",
+						Namespace: "ns",
+					},
+					Spec: networking.NetworkPolicySpec{
+						PodSelector: metav1.LabelSelector{},
+						PolicyTypes: []networking.PolicyType{networking.PolicyTypeEgress},
+						Egress: []networking.NetworkPolicyEgressRule{
+							{
+								To: []networking.NetworkPolicyPeer{
+									{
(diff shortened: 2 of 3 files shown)


.github/workflows/pr-tests.yaml
```diff
@@ -55,9 +55,9 @@ jobs:
           check-latest: true
           cache-dependency-path: "**/go.sum"
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v2
+        uses: docker/setup-qemu-action@v3
```
parlakisik, line 58: what is the reason for bumping docker versions .

oliviassss: @v2 is broken on current GitHub runners due to Node.js 20 deprecation and runner Docker changes. @v3 is the fix.
I saw the gh actions failing on this thus bumped up the version.

pkg/resolvers/endpoints_test.go
```diff
@@ -495,6 +495,45 @@ func TestEndpointsResolver_Resolve(t *testing.T) {
 				{PodIP: "1.0.0.3", Name: "pod3", Namespace: "ns"},
 			},
 		},
+		{
+			name: "egress ipBlock CIDR with nil Protocol defaults to TCP",
```
jaydeokar, line 499: nil port also defaults to TCP.. (all ports, TCP default) 
Can we add a TC for this ?

.github/workflows/pr-tests.yaml
parlakisik: Got it. Can you add this information to the PR description? Right now, you are fixing two problems. It is good to know this when looking back at previous commits in case of any issues.

parlakisik (dismissed):

pkg/resolvers/endpoints_test.go
oliviassss: added for nil port case.

parlakisik (approved):
lgtm

PR #243, opened 2026-07-21
Fix: nil protocol dereference in getportlist

Pavani-Panakanti:
### What this PR does

Fixes a nil-pointer dereference panic in getPortList (pkg/resolvers/endpoints.go) that crash-loops the controller when a policy rule contains a port entry with protocol omitted.

 ### Issue
  
NetworkPolicyPort.Protocol is *corev1.Protocol and optional — per the K8s API spec, a nil protocol should default to TCP. PR #237 (released in v1.1.9) added a nil guard in convertToPolicyInfoPortForCIDRs, but missed getPortList, which still dereferences *port.Protocol unguarded:
```
  portVal, _, err := k8s.LookupContainerPortAndName(&pod, *port.Port, *port.Protocol)
  //                                            panics when Protocol is nil ^^^^^^^^^^

  getPortList is on the named-port resolution path, reached from both:
  - egress: computeEgressEndpoints → getPortList (per destination pod)
  - ingress: resolveIngressPorts → getIngressRulesPorts → getPortList
```
  Any rule with a port entry whose protocol is nil crashes the controller with SIGSEGV, causing CrashLoopBackOff. Because the offending policy persists, the controller crash-loops indefinitely. 
Note: native networking.k8s.io/v1 NetworkPolicy objects get protocol: TCP defaulted by the API server at admission, but ApplicationNetworkPolicy (CRD) has no such defaulting — nil Protocol reaches the resolver directly.
```
  panic: runtime error: invalid memory address or nil pointer dereference

  goroutine 292 [running]:
  ...(*defaultEndpointsResolver).getPortList(...)
        /workspace/pkg/resolvers/endpoints.go:360 +0x17a
  ...(*defaultEndpointsResolver).Resolve.func2()
        /workspace/pkg/resolvers/endpoints.go:64 +0x69
```
  ### Fix

  Applies the same nil-guard pattern as PR #237:

  1. getPortList — default protocol to TCP when port.Protocol is nil, pass the defaulted value to LookupContainerPortAndName, and emit Protocol: &protocol (always non-nil) in the output policyinfo.Port so nil no longer propagates downstream.
  2. getMatchingServiceClusterIPs — hardening: hoisted the existing npProto TCP-default above the if npPort.Port != nil block and emit Protocol: &npProto in output. The dereference here was already guarded (no panic), but the output re-propagated the caller's nil Protocol into PolicyEndpoints.

  A full-repo scan for unguarded *.Protocol / *.Port / *.EndPort dereferences confirmed these are the last remaining sites — all other dereferences (dedupPorts, isNumericPortAllowed, equality/hashing code, ANP/CNP resolvers) are already nil-guarded.

### Testing

  Unit tests (all passing, incl. -race)
  
  - TestEndpointsResolver_getPortList (new, 8 table cases): nil Protocol + named port, nil Protocol + numeric port, nil Protocol + nil Port
  (protocol-only), explicit TCP/UDP controls, mixed list (named+TCP followed by numeric nil-protocol), multiple nil-Protocol entries. Verified RED on
  unfixed code (panics at endpoints.go:360 with the exact production stack trace) → GREEN with the fix.
  - TestEndpointsResolver_convertToPolicyInfoPortForCIDRs (new, 7 table cases): pins the PR #237 fix with direct regression tests (previously only covered
  indirectly).
  - TestGetMatchingServiceClusterIPs_NilProtocolDefaultsTCP (new): asserts service ClusterIP endpoint ports carry non-nil TCP-defaulted Protocol when the
  NP port omits protocol, including the protocol-only entry (nil Port + nil Protocol) case. Verified RED on unfixed code → GREEN with the fix.
  - Full existing suite passes: go build ./..., go vet ./..., gofmt clean, go test ./..., go test -race.

  Cluster verification (EKS, NPC on dataplane)

  Built two images: one from the v1.1.9 tag (unfixed) and one from this branch, and ran before/after on a live cluster with fixture pods (backend exposing named port http, web exposing named metrics + numeric 8443, scraper):

```  ┌───────────────────────────────────────────────────────────┬───────────────────────────────────────────────────────────────┬───────────────────────┐
  │          Repro policy (ApplicationNetworkPolicy)          │                            v1.1.9                             │       This fix        │
  ├───────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────┼───────────────────────┤
  │ Egress, named port, protocol omitted                      │ ❌ panic at endpoints.go:360 via Resolve.func2 (matches       │ ✅ reconciles, 0      │
  │                                                           │ production stack)                                             │ restarts              │
  ├───────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────┼───────────────────────┤
  │ Ingress, named port, protocol omitted                     │ ❌ panic at endpoints.go:360 via Resolve.func1 →              │ ✅ reconciles, 0      │
  │                                                           │ getIngressRulesPorts                                          │ restarts              │
  ├───────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────┼───────────────────────┤
  │ Mixed rule (named+TCP entry, numeric entry with protocol  │ ❌ panic at endpoints.go:360                                  │ ✅ reconciles, 0      │
  │ omitted)                                                  │                                                               │ restarts              │
  ├───────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────┼───────────────────────┤
  │ Regression: NP egress ipBlock CIDR, protocol omitted (PR  │ ✅ no panic                                                   │ ✅ reconciles, 0      │
  │ #237 path)                                                │                                                               │ restarts              │
  └───────────────────────────────────────────────────────────┴───────────────────────────────────────────────────────────────┴───────────────────────┘
```
  With the fixed image, all four policies were also applied simultaneously — zero restarts over the observation window, logs clean of panics, and PolicyEndpoints show the TCP-defaulted protocol on resolved ports.

  Behavior changes

  - Port entries with omitted protocol now resolve with TCP (per K8s spec) instead of crashing.
  - policyinfo.Port.Protocol emitted by getPortList and getMatchingServiceClusterIPs is now always non-nil (TCP-defaulted). This matches what the PolicyEndpoints CRD already defaults on write (default: TCP), and removes a latent nil-vs-TCP semantic-equality diff on protocol-only entries.

```release-note

```

### DIFF
--- a/pkg/resolvers/endpoints.go
+++ b/pkg/resolvers/endpoints.go
@@ -355,9 +355,13 @@ func (r *defaultEndpointsResolver) getIngressRulesPorts(ctx context.Context, pol
 func (r *defaultEndpointsResolver) getPortList(pod corev1.Pod, ports []networking.NetworkPolicyPort) []policyinfo.Port {
 	var portList []policyinfo.Port
 	for _, port := range ports {
+		protocol := corev1.ProtocolTCP
+		if port.Protocol != nil {
+			protocol = *port.Protocol
+		}
 		var portPtr *int32
 		if port.Port != nil {
-			portVal, _, err := k8s.LookupContainerPortAndName(&pod, *port.Port, *port.Protocol)
+			portVal, _, err := k8s.LookupContainerPortAndName(&pod, *port.Port, protocol)
 			if err != nil {
 				// Isolate the pod for the port if we are unable to resolve the named port
 				r.logger.Info("Unable to lookup container port", "pod", k8s.NamespacedName(&pod),
@@ -367,7 +371,7 @@ func (r *defaultEndpointsResolver) getPortList(pod corev1.Pod, ports []networkin
 			portPtr = &portVal
 		}
 		portList = append(portList, policyinfo.Port{
-			Protocol: port.Protocol,
+			Protocol: &protocol,
 			Port:     portPtr,
 			EndPort:  port.EndPort,
 		})
@@ -519,12 +523,12 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
 
 		var portList []policyinfo.Port
 		for _, npPort := range npPorts {
+			npProto := corev1.ProtocolTCP
+			if npPort.Protocol != nil {
+				npProto = *npPort.Protocol
+			}
 			var portPtr *int32
 			if npPort.Port != nil {
-				npProto := corev1.ProtocolTCP
-				if npPort.Protocol != nil {
-					npProto = *npPort.Protocol
-				}
 				portVal, err := r.getMatchingServicePort(svc, npPort.Port, npProto, getAllPods)
 				if err != nil {
 					r.logger.Info("Unable to lookup service port", "err", err)
@@ -544,7 +548,7 @@ func (r *defaultEndpointsResolver) getMatchingServiceClusterIPs(ctx context.Cont
 				portPtr = &portVal
 			}
 			portList = append(portList, policyinfo.Port{
-				Protocol: npPort.Protocol,
+				Protocol: &npProto,
 				Port:     portPtr,
 				EndPort:  npPort.EndPort,
 			})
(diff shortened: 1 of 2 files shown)


srikanthjg (approved):
lgtm

Issue #224, opened 2026-02-24
Network Policy not applied when podSelector labels differ from selector in target service


der-eismann:
**What happened**:
We started to use network policies like described in [the guide](https://docs.aws.amazon.com/eks/latest/userguide/cni-network-policy.html). We locked down a pod to only allow access to certain other pods and while most of the policies worked fine, one outside pod was unreachable even though we allowed it.
We enabled [network policy logs](https://docs.aws.amazon.com/eks/latest/userguide/network-policies-troubleshooting.html#network-policies-troubleshooting-flowlogs) and weirdly enough we could see that the working pods were accessed through their pod IP while the blocked one was accessed through the service IP.

The network policy:
```yaml
    - to:
        - podSelector:
            matchLabels:
              app.kubernetes.io/name: service-a
      ports:
        - port: 8080
    - to:
        - podSelector:
            matchLabels:
              app.kubernetes.io/name: service-b
      ports:
        - port: 8080
```

```
Working connection to service-a using pod IP:
{"level":"debug","ts":"2026-02-23T14:19:09.129Z","caller":"runtime/asm_amd64.s:1693","msg":"Flow Info: Src IP: 172.22.27.90 Src Port: 59838 Dest IP: 172.22.26.253 Dest Port: 8080 Proto TCP Verdict ACCEPT Direction ingress, Tier DEFAULT"}
Broken connection to service-b using service IP:
{"level":"info","ts":"2026-02-23T14:09:22.250Z","caller":"runtime/asm_amd64.s:1693","msg":"Flow Info: Src IP: 172.22.27.90 Src Port: 41398 Dest IP: 10.10.134.38 Dest Port: 80 Proto TCP Verdict DENY Direction egress, Tier NETWORK_POLICY"}
```

We ensured that `service-b` has the label from the selector:
```yaml
  labels:
    app: service-b
    app.kubernetes.io/instance: service-b-default
    app.kubernetes.io/name: service-b
    app.kubernetes.io/version: default
```

 BUT: The k8s service for `service-b` used a different selector:
```yaml
spec:
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: 8080
  selector:
    app: service-b
```

After changing the network policy to use the same selector it suddenly started to work and access to `service-b` was allowed:
```yaml
    - to:
        - podSelector:
            matchLabels:
              app: service-b
      ports:
        - port: 8080
```

**Attach logs**

**What you expected to happen**:

[...]

PR #252, opened 2026-08-07
Document podSelector limitation from #224

cdirubbio:
**What type of PR is this?** documentation

**Which issue does this PR fix**: #224

**What does this PR do / Why do we need it**: Document unsupported behavior with selecting label in NP that is not selected on in Service. This will lead to pod IPs allowed, service IPs not allowed.

**Automation added to e2e**:

```release-note

```


jaydeokar (approved):

jaydeokar:
lgtm

PR #260, opened 2026-09-02
fix: make PolicyEndpoint rule-identity hashes injective

nakshah87:
## Problem

`getEndpointInfoKey` / `getClusterEndpointInfoKey` produce the digest that `processExistingPolicyEndpoints` / `processExistingClusterPolicyEndpoints` use as their **only** equality test for reusing a rule already stored on a `(Cluster)PolicyEndpoint`:

```go
ruleKey := m.getEndpointInfoKey(ingRule)
if _, exists := ingressEndpointsMap[ruleKey]; exists {
    ingEndpointList = append(ingEndpointList, ingRule) // keep STORED rule
    delete(ingressEndpointsMap, ruleKey)               // drop DESIRED rule
}
```

On a digest collision the controller keeps the stale rule and discards the desired one. The object is then never rewritten — `resourceVersion` freezes while the policy's `generation` advances — so the node agent keeps enforcing a revoked rule. Nothing errors, nothing logs.

Both encodings were non-injective, for two unrelated reasons:

1. **Separator-free concatenation** (namespaced path). Fields ran together and ports were decimal text, so `{port: 1, endPort: 23}` and `{port: 123}` hash identically. Narrowing `1-23` → `123` is silently ignored; so is widening back. Same class for `CIDR`/`Except` and `{80, 8080}` vs `{8080, 80}`.

2. **`string(rune(port))`** (cluster path) writes the code point, not the decimal text. Every value in the UTF-16 surrogate block **U+D800–U+DFFF (55296–57343)** is an invalid rune and encodes to U+FFFD, so all **2048** ports there hash identically.

These are different bugs. The cluster path does *not* share the digit collision — code points are fixed-width, hence prefix-free — it has its own, wider one.

## Fix

Every field is written delimited and self-describing, ports in decimal:

```go
fmt.Fprintf(hasher, "proto=%s|port=%s|endPort=%s|",
    protocolKeyPart(port.Protocol), int32PtrKeyPart(port.Port), int32PtrKeyPart(port.EndPort))
```

Helpers emit `"nil"` for absent pointers, keeping a missing port distinct from port `0`. Previously an absent field wrote nothing, itself a collision source.

**Defence in depth:** the digest now only *selects a candidate*. Both reuse gates also require the rules to actually match, so correctness no longer depends on the hash being perfect:

```go
if desired, exists := ingressEndpointsMap[ruleKey]; exists && equality.Semantic.DeepEqual(desired, ingRule) {
```

## Upgrade: no migration needed

The digest now covers every field rather than a subset, so stale objects are rewritten on the first reconcile after upgrade: the informer's initial LIST fires `Create` events, `enqueueRequestForPolicyEvent.Create` enqueues unconditionally, `SetupWithManager` registers no predicates, and `reconcile` calls `reconcileNetworkPolicy` unconditionally.

Caveat: this is tied to controller **start**. `Update` short-circuits on unchanged spec, so a long-running controller won't re-derive keys for an untouched policy. Deleting `PolicyEndpoint`s to force a rebuild is *not* a safe alternative — it opens an enforcement gap.

## Second, unrelated fix: chart ClusterRole

The chart's `ClusterRole` was never regenerated after `ClusterNetworkPolicy` / `ApplicationNetworkPolicy` landed. It grants only `policyendpoints`, while `config/rbac/role.yaml` and the markers at `internal/controllers/policy_controller.go:90-95` require `clusternetworkpolicies`, `clusterpolicyendpoints`, `applicationnetworkpolicies` and their `/status`, `/finalizers`.

Failure mode is silent and total: those informers get `forbidden`, `WaitForCacheSync` never completes, and the manager starts **no controllers at all** — while the Deployment reports Ready and `kubectl rollout status` succeeds.

Not fixed here: `scripts/deploy-controller-on-dataplane.sh` passes `--set useExistingClusterRole=true`, but nothing in `charts/` reads it. The chart always creates its own `ClusterRole` and binds to it, shadowing the managed `eks:network-policy-controller` role that already has every rule. Follow-up: implement the value or drop the flag.

## Testing

**Unit** — `pkg/policyendpoints/hash_collision_test.go`, all failing before this change:

| Test | Guards |
|---|---|
| `..._portRangeVsSinglePort_noCollision` (both paths) | `{1, endPort 23}` vs `{123}` |
| `Test_getEndpointInfoKey_deterministicAndDistinct` | determinism; absent port vs `0`; `{cidr /2, except 4}` vs `{cidr /24}` |
| `Test_getClusterEndpointInfoKey_distinctPorts` | `80` vs `443`; `Accept` vs `Deny` |
| `..._surrogateBlockPorts_noCollision` | `55296` vs `55297`/`57000`/`57343`, and vs `40000` |

Old-encoding collisions confirmed by re-implementing it standalone:

```
cluster OLD: 55296 vs 55297 collide=true
cluster OLD: 55296 vs 57000 collide=true
cluster OLD: 55296 vs 57343 collide=true
cluster OLD: 55296 vs 40000 collide=false   <- control, correctly distinct
ns      OLD: {1,23} vs {123}  collide=true
```

`go build ./...`, `go vet ./pkg/...`, `gofmt -l`, `go test ./...`, `helm template` all clean.

**Live EKS dev cluster** — this branch built into an image and deployed to the dataplane via `scripts/deploy-controller-on-dataplane.sh`.

*Baseline, unpatched:* editing the policy's port block between `{TCP 123}` and `{TCP 1-23}` froze its `PolicyEndpoint` in **both** directions — the stale rule was reused every time. At the point captured below the policy declared `{TCP 1-23}` at `generation=5`, while the `PolicyEndpoint` was still holding the superseded `{TCP port 123}` at `rv=15070982`, unchanged 5 min later: five accepted spec versions, zero downstream writes. A control policy on port `124` (no colliding partner) propagated in <30s, confirming the controller was otherwise healthy.

*Self-heal on start:* with that stale object left in place, the patched controller was started. No policy edit — `generation` stayed 5:

| | rv | stored rule |
|---|---|---|
| before | `15070982` | `{TCP port 123}` |
| after start | `15134169` | `{TCP port 1, endPort 23}` ✅ |

*Regression checks*, each previously frozen indefinitely, now applied within 10s:

| Change | Result |
|---|---|
| narrow `1-23` → `123` | rv `15134544` |
| widen `123` → `1-23` | rv `15134836` |
| CNP port `55296` → `57000` (both U+FFFD before) | rv `15135035` |

**End-to-end traffic enforcement.** The above confirms the controller writes the right object; this confirms packets actually follow it. Two busybox pods in namespace `nptraffic` — `server` listening on TCP 22 and 123, `client` probing with `nc -w 3` — under a policy admitting only `172.31.0.0/16`. Same chart (RBAC fix included) for both runs, so the only variable is the controller binary.

| Policy declares | Controller | `PolicyEndpoint` stores | port 22 | port 123 |
|---|---|---|---|---|
| `TCP 1-23` | unpatched | `1-23` | ALLOW | DENY |
| `TCP 123` | unpatched | `1-23` (frozen, rv `15367688`) | **ALLOW** ⚠️ | **DENY** ⚠️ |
| `TCP 123` | patched, *no policy edit* | `123` (rv `15368576`) | DENY ✅ | ALLOW ✅ |
| `TCP 1-23` | patched | `1-23` (rv `15368767`) | ALLOW ✅ | DENY ✅ |

Row 2 is the impact in one line: `generation` had advanced to 2 declaring **only** port 123, yet port 22 stayed reachable for over 2 minutes — a revoked port still admitting traffic — while port 123, explicitly allowed, was dropped. Both a policy bypass and an availability break, from one digest collision.

Row 3 is self-heal at the traffic layer: no policy edit, only the patched controller starting.

Denies were confirmed to be eBPF **drops** rather than connection refusals, by timing (~1.2s of each figure is fixed `kubectl exec` overhead):

| probe | policy | listener | elapsed |
|---|---|---|---|
| port 22 | allows | yes | 1245ms — connects |
| port 21 | allows | **no** | 1171ms — instant RST, refused |
| port 123 | denies | yes | 4199ms — full `nc -w 3` timeout, dropped |

Port 21 is the control that matters: permitted by the policy but with nothing listening, it refuses immediately. So the extra 3s on port 123 is the dataplane discarding packets, not a missing listener.

*Chart RBAC:* before that commit the deploy script yielded a Ready 2/2 Deployment that reconciled nothing — the stale object stayed at `15070982` even under the patched controller. Granting the three resources dropped `forbidden` lines to zero and reconciliation resumed. That's how the gap was found.

## Risk

All existing `(Cluster)PolicyEndpoint`s get keys recomputed on first reconcile, so any whose stored rules disagree with their policy are rewritten once. That's the intended correction, not churn — the digest is a pure function of existing fields, so objects already in agreement hash consistently and are left untouched.

### DIFF
--- a/pkg/policyendpoints/hash_collision_test.go
+++ b/pkg/policyendpoints/hash_collision_test.go
@@ -9,8 +9,8 @@ import (
 	policyinfo "github.com/aws/amazon-network-policy-controller-k8s/api/v1alpha1"
 )
 
-// Regression tests for V2339293193: a separator-free rule-identity hash let
-// the controller treat two semantically different rules as one. Narrowing a
+// Regression tests for a rule-identity hash collision: a separator-free hash
+// let the controller treat two semantically different rules as one. Narrowing a
 // NetworkPolicy from the port range 1-23 to the single port 123 produced the
 // same digest, so the controller kept the revoked range and dropped the newly
 // declared port.
--- a/pkg/policyendpoints/hash_collision_test.go
+++ b/pkg/policyendpoints/hash_collision_test.go
@@ -1,0 +1,133 @@
+package policyendpoints
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+
+	policyinfo "github.com/aws/amazon-network-policy-controller-k8s/api/v1alpha1"
+)
+
+// Regression tests for V2339293193: a separator-free rule-identity hash let
+// the controller treat two semantically different rules as one. Narrowing a
+// NetworkPolicy from the port range 1-23 to the single port 123 produced the
+// same digest, so the controller kept the revoked range and dropped the newly
+// declared port.
+
+func i32(v int32) *int32 { return &v }
+
+func Test_getEndpointInfoKey_portRangeVsSinglePort_noCollision(t *testing.T) {
+	m := &policyEndpointsManager{}
+	tcp := corev1.ProtocolTCP
+
+	// {TCP, port 1, endPort 23} -- the port RANGE 1-23
+	rangeRule := policyinfo.EndpointInfo{
+		CIDR:  "10.0.0.0/24",
+		Ports: []policyinfo.Port{{Protocol: &tcp, Port: i32(1), EndPort: i32(23)}},
+	}
+	// {TCP, port 123} -- a single port whose digits concatenate to the range's
+	singleRule := policyinfo.EndpointInfo{
+		CIDR:  "10.0.0.0/24",
+		Ports: []policyinfo.Port{{Protocol: &tcp, Port: i32(123)}},
+	}
+
+	assert.NotEqual(t, m.getEndpointInfoKey(rangeRule), m.getEndpointInfoKey(singleRule),
+		"range 1-23 and single port 123 must not share a hash key")
+}
+
+func Test_getEndpointInfoKey_deterministicAndDistinct(t *testing.T) {
+	m := &policyEndpointsManager{}
+	tcp := corev1.ProtocolTCP
+
+	base := policyinfo.EndpointInfo{
+		CIDR:  "10.0.0.0/24",
+		Ports: []policyinfo.Port{{Protocol: &tcp, Port: i32(80)}},
+	}
+	// same input -> same key
+	assert.Equal(t, m.getEndpointInfoKey(base), m.getEndpointInfoKey(base))
+
+	// a nil port must not collide with port 0 (the "nil" sentinel guards this)
+	nilPort := policyinfo.EndpointInfo{CIDR: "10.0.0.0/24", Ports: []policyinfo.Port{{Protocol: &tcp}}}
+	zeroPort := policyinfo.EndpointInfo{CIDR: "10.0.0.0/24", Ports: []policyinfo.Port{{Protocol: &tcp, Port: i32(0)}}}
+	assert.NotEqual(t, m.getEndpointInfoKey(nilPort), m.getEndpointInfoKey(zeroPort),
+		"absent port and port 0 must be distinguishable")
+
+	// CIDR boundary that a separator-free hash would blur:
+	// {cidr 10.0.0.0/2, except 4} vs {cidr 10.0.0.0/24}
+	a := policyinfo.EndpointInfo{CIDR: "10.0.0.0/2", Except: []policyinfo.NetworkAddress{"4"}}
+	b := policyinfo.EndpointInfo{CIDR: "10.0.0.0/24"}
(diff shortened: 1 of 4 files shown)


pkg/policyendpoints/hash_collision_test.go
```diff
@@ -0,0 +1,133 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+
+	policyinfo "github.com/aws/amazon-network-policy-controller-k8s/api/v1alpha1"
+)
+
+// Regression tests for V2339293193: a separator-free rule-identity hash let
```
cdirubbio: remove ticket id

nakshah87: Done

zac-nixon (approved):

parlakisik (approved):
lgtm

PR #261, opened 2026-09-02
Make policy endpoint hash keys unambiguous

haouc:
## What

Endpoint hash keys are used as rule identity in the chunking code, so the key derivation has to be injective. It was not: fields were concatenated without a delimiter, and ports were converted to a string via a rune conversion, which is lossy over part of the input range.

Two distinct rules could therefore derive the same key. A key match was taken as proof of rule equality, and the subsequent change check compared only list lengths, so a rule edited in place could be dropped — leaving the stored `PolicyEndpoint` / `ClusterPolicyEndpoint` holding the previous value with no update written.

## Changes

- Add `hashKeyWriter`, which tags every field, length-prefixes every variable-length value, and renders integers as decimal digits.
- Rewrite `getClusterEndpointInfoKey` and `getEndpointInfoKey` to use it, covering both the ClusterNetworkPolicy and the namespaced NetworkPolicy / ANP paths.
- On a key match, keep the desired rule rather than the stored one, so correctness no longer depends on the key being collision-free.
- Compare list contents with `equality.Semantic.DeepEqual` rather than comparing lengths when deciding whether an endpoint object changed. An in-place replacement leaves the count unchanged, so the length check alone would treat it as no change and never persist it.

## Upgrade behavior

Existing objects converge on the first reconcile after upgrade: a stale stored rule no longer key-matches the desired rule, so it is replaced. Rules that are already correct still key-match and are left alone. No migration step is required.

## Testing

New unit tests in `pkg/policyendpoints/hashkey_test.go`:

| Test | Covers |
|---|---|
| `TestClusterKeyInjectiveOverPortSpace` | every port in 1..65535 maps to a distinct key |
| `TestClusterKeyHighPortsDiffer` | port pairs that previously shared a key |
| `TestKeyFieldsCannotAlias` | one field's value cannot absorb the next |
| `TestNamespacedKeyPortRangesCannotAlias` | adjacent numeric fields do not run together; CIDR does not absorb an `Except` entry |
| `TestInPlacePortEditReachesCPE` | an in-place port edit reaches the persisted CPE via `computeClusterPolicyEndpoints` |

```
--- PASS: TestClusterKeyInjectiveOverPortSpace (0.08s)
--- PASS: TestClusterKeyHighPortsDiffer (0.00s)
--- PASS: TestKeyFieldsCannotAlias (0.00s)
--- PASS: TestNamespacedKeyPortRangesCannotAlias (0.00s)
--- PASS: TestInPlacePortEditReachesCPE (0.00s)
```

All five fail on unmodified source and pass with this change. `go test ./...` and `go vet ./...` are clean; the existing `pkg/policyendpoints` and `internal/controllers` suites still pass.

## Notes

Not addressed here: the pre-existing `TODO` on `getEndpointInfoKey` about port-slice ordering producing different keys for semantically equal rules. With the content comparison in place that now causes a converging rewrite rather than stale state, so it is left for a separate change.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/pkg/policyendpoints/clusterpolicyendpoints_chunking.go
+++ b/pkg/policyendpoints/clusterpolicyendpoints_chunking.go
@@ -1,9 +1,6 @@
 package policyendpoints
 
 import (
-	"crypto/sha256"
-	"encoding/hex"
-
 	"github.com/samber/lo"
 	"golang.org/x/exp/maps"
 	"k8s.io/apimachinery/pkg/api/equality"
@@ -57,11 +54,12 @@ func (m *policyEndpointsManager) processExistingClusterPolicyEndpoints(
 	// Process existing CPEs (following the same logic as regular PE processing)
 	for i := range existingCPEs {
 		// Filter ingress rules - keep only those that exist in the new rules
+		// Keep the desired copy; a key match need not imply equality.
 		ingEndpointList := make([]policyinfo.ClusterEndpointInfo, 0, len(existingCPEs[i].Spec.Ingress))
 		for _, ingRule := range existingCPEs[i].Spec.Ingress {
 			ruleKey := m.getClusterEndpointInfoKey(ingRule)
-			if _, exists := ingressEndpointsMap[ruleKey]; exists {
-				ingEndpointList = append(ingEndpointList, ingRule)
+			if desired, exists := ingressEndpointsMap[ruleKey]; exists {
+				ingEndpointList = append(ingEndpointList, *desired.DeepCopy())
 				delete(ingressEndpointsMap, ruleKey)
 			}
 		}
@@ -70,8 +68,8 @@ func (m *policyEndpointsManager) processExistingClusterPolicyEndpoints(
 		egEndpointList := make([]policyinfo.ClusterEndpointInfo, 0, len(existingCPEs[i].Spec.Egress))
 		for _, egRule := range existingCPEs[i].Spec.Egress {
 			ruleKey := m.getClusterEndpointInfoKey(egRule)
-			if _, exists := egressEndpointsMap[ruleKey]; exists {
-				egEndpointList = append(egEndpointList, egRule)
+			if desired, exists := egressEndpointsMap[ruleKey]; exists {
+				egEndpointList = append(egEndpointList, *desired.DeepCopy())
 				delete(egressEndpointsMap, ruleKey)
 			}
 		}
@@ -102,11 +100,12 @@ func (m *policyEndpointsManager) processExistingClusterPolicyEndpoints(
 			existingCPEs[i].Spec.Egress = egEndpointList
 			existingCPEs[i].Spec.PodSelectorEndpoints = podSelectorEndpointList
 			potentialDeletes = append(potentialDeletes, existingCPEs[i])
-		} else if len(existingCPEs[i].Spec.Ingress) != len(ingEndpointList) ||
-			len(existingCPEs[i].Spec.Egress) != len(egEndpointList) ||
-			len(existingCPEs[i].Spec.PodSelectorEndpoints) != len(podSelectorEndpointList) ||
+		} else if !equality.Semantic.DeepEqual(existingCPEs[i].Spec.Ingress, ingEndpointList) ||
+			!equality.Semantic.DeepEqual(existingCPEs[i].Spec.Egress, egEndpointList) ||
+			!equality.Semantic.DeepEqual(existingCPEs[i].Spec.PodSelectorEndpoints, podSelectorEndpointList) ||
 			tierChanged || priorityChanged || subjectChanged {
-			// CPE has changed - update it
+			// CPE has changed - update it. Content, not length: an in-place
+			// edit keeps the count.
 			existingCPEs[i].Spec.Ingress = ingEndpointList
 			existingCPEs[i].Spec.Egress = egEndpointList
 			existingCPEs[i].Spec.PodSelectorEndpoints = podSelectorEndpointList
@@ -279,20 +278,10 @@ func (m *policyEndpointsManager) getClusterEndpointInfoFromHashes(hashes []strin
 
 // getClusterEndpointInfoKey generates a hash key for ClusterEndpointInfo
 func (m *policyEndpointsManager) getClusterEndpointInfoKey(info policyinfo.ClusterEndpointInfo) string {
-	hasher := sha256.New()
-	hasher.Write([]byte(string(info.CIDR)))
-	hasher.Write([]byte(string(info.DomainName)))
-	hasher.Write([]byte(string(info.Action)))
-	for _, port := range info.Ports {
-		if port.Protocol != nil {
-			hasher.Write([]byte(string(*port.Protocol)))
-		}
-		if port.Port != nil {
-			hasher.Write([]byte(string(rune(*port.Port))))
-		}
-		if port.EndPort != nil {
-			hasher.Write([]byte(string(rune(*port.EndPort))))
-		}
-	}
-	return hex.EncodeToString(hasher.Sum(nil))
+	w := newHashKeyWriter()
+	w.str("cidr", string(info.CIDR))
+	w.str("domain", string(info.DomainName))
+	w.str("action", string(info.Action))
+	w.ports(info.Ports)
+	return w.sum()
 }
--- a/pkg/policyendpoints/hashkey.go
+++ b/pkg/policyendpoints/hashkey.go
@@ -1,0 +1,118 @@
+package policyendpoints
+
+import (
+	"cmp"
+	"crypto/sha256"
+	"encoding/hex"
+	"slices"
+	"strconv"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+
+	policyinfo "github.com/aws/amazon-network-policy-controller-k8s/api/v1alpha1"
+)
+
+// hashKeyWriter builds injective hash input for endpoint keys: fields are tagged
+// and byte-length-prefixed, integers are decimal. A cluster rule encodes as
+//
+//	cidr=13:10.200.0.0/16;domain=0:;action=4:Deny;ports=1[proto=3:TCP;port=80;endport=nil;,];
+type hashKeyWriter struct {
+	buf []byte
+}
+
+func newHashKeyWriter() *hashKeyWriter {
+	return &hashKeyWriter{buf: make([]byte, 0, 128)}
+}
+
+func (w *hashKeyWriter) str(tag, val string) {
+	w.buf = append(w.buf, tag...)
+	w.buf = append(w.buf, '=')
+	w.buf = strconv.AppendInt(w.buf, int64(len(val)), 10)
+	w.buf = append(w.buf, ':')
+	w.buf = append(w.buf, val...)
+	w.buf = append(w.buf, ';')
+}
+
+func (w *hashKeyWriter) optInt32(tag string, val *int32) {
+	w.buf = append(w.buf, tag...)
+	w.buf = append(w.buf, '=')
+	if val == nil {
+		w.buf = append(w.buf, "nil;"...)
+		return
+	}
+	w.buf = strconv.AppendInt(w.buf, int64(*val), 10)
+	w.buf = append(w.buf, ';')
+}
+
+func (w *hashKeyWriter) strs(tag string, vals []policyinfo.NetworkAddress) {
+	sorted := vals
+	if !slices.IsSorted(vals) {
+		sorted = slices.Clone(vals)
+		slices.Sort(sorted)
+	}
+
+	w.buf = append(w.buf, tag...)
+	w.buf = append(w.buf, '=')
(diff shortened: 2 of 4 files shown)


haouc:
Addressed all five actionable items from Copilot. Summary of what changed and what I verified.

**1. Reusing the desired object on a key match (`clusterpolicyendpoints_chunking.go`, `manager.go`)**

Real aliasing, now fixed with `*desired.DeepCopy()` in both the cluster and namespaced paths.

Worth noting the path is narrower than it looks: when the retained rules are content-identical, `processExisting*PolicyEndpoints` falls into the final `else` ("unchanged — keep as is") branch, which never assigns the newly built list, so the desired object is never installed. It is only installed when the spec is genuinely rewritten (tier / priority / subject / pod-selector change). My first version of the test missed this and passed even with the `DeepCopy` removed. After driving it through a tier change, the negative control fails as it should:

```
--- FAIL: TestSubstitutedRuleIsIsolated
    hashkey_test.go:176: mutating the persisted rule reached the desired rule
```

Covered by `TestSubstitutedRuleIsIsolated`.

**2. `equality.Semantic.DeepEqual` for change detection**

Correct here, and deliberately not `reflect.DeepEqual`. `equality.Semantic` treats a nil slice and an empty slice as equal:

| | nil vs `[]T{}` |
|---|---|
| `reflect.DeepEqual` | `false` |
| `equality.Semantic.DeepEqual` | `true` |

The filtered lists are built with `make([]T, 0, n)`, so they are non-nil even when empty, while a stored spec that has never had rules deserializes as nil. Under `reflect.DeepEqual` every such object would be flagged changed on every reconcile — an unnecessary write per object per loop. Verified no spurious update is produced for a spec with nil ingress.

**3. Deterministic port ordering**

Implemented rather than documented — the key now hashes a canonical ordering, so a permuted `Ports` slice yields the same key. This also closes the pre-existing `TODO` above `getEndpointInfoKey`, which is removed. Same treatment for `EndpointInfo.Except`, which had the identical problem in the same function. Both sort a copy, so the caller's slice is never reordered (asserted in `TestKeyIgnoresFieldOrder`).

**4. nil / empty / boundary coverage**

`TestKeyNilAndBoundaryPortFields` enumerates the full cross product of `Protocol` ∈ {nil, TCP, UDP} × `Port` ∈ {nil, 1, 65535} × `EndPort` ∈ {nil, 1, 65535} — 27 combinations, asserting all 27 keys are distinct and each is stable across repeated calls. It also asserts nil and empty `Ports` are equivalent, and that neither collides with a single-port rule.

**5. `go vet` / `go test ./...`**

Both clean locally, and CI runs them on this branch.

---

Non-blocking items:

**(b) Benchmark.** One correction to the premise: the new code does not build a string key — it appends into a byte buffer and hashes once. Measured against the previous implementation, same machine, 6×20000 iterations:

| | before | after |
|---|---|---|
| 1 port | 252 ns/op, 160 B, 3 allocs | 389 ns/op, 256 B, 3 allocs |
| 16 ports | 742 ns/op, 160 B, 3 allocs | 1698 ns/op, 1152 B, 4 allocs |

So ~1.5× on the common single-port rule and ~2.3× on a 16-port rule. That is the honest cost of an injective encoding: the old version wrote 1–3 bytes per port, the new one writes a tagged, length-prefixed field. My first cut used `fmt.Fprintf` and was 5–9× slower at 13–90 allocs; switching to buffer appends with `strconv.AppendInt`, `slices.SortFunc` over a generic comparator, an already-sorted fast path that skips the copy entirely, and a single `slices.Grow` brought it to the above. At a few hundred ns per rule this is not measurable against a reconcile. `BenchmarkClusterEndpointInfoKey` is committed so the number is trackable; the old implementation is not committed alongside it.

**(a) Invariant/upgrade comment.** The injectivity invariant is documented on `hashKeyWriter`; I have kept it to that deliberately. On upgrade, keys change for every existing object, so the first reconcile after rollout re-derives them — the existing chunking logic handles this as an ordinary rule-set change, and no migration is needed.

**(c) Port-ordering test.** Superseded by (3) — since ordering is now normalized rather than left as-is, `TestKeyIgnoresFieldOrder` asserts the desired behaviour directly instead of documenting a divergence.

haouc:
Second round addressed. Three items were real gaps and are now fixed; the rest I could answer from the code, including one that rests on a false premise.

## Real gaps, now closed

**Duplicate stored rules mapping to the same key.** Good catch — this was untested, and the answer is better than the review feared. Duplicates collapse in a **single** reconcile, not one-per-reconcile:

| scenario | result |
|---|---|
| one object holds rule X twice | retained once; no desired rule left unplaced |
| two objects each hold rule X | retained in the first; the second is emptied into `potentialDeletes` |

That is the correct outcome: the emptied object is a genuine delete candidate, and no desired rule is dropped. Locked in by `TestDuplicateStoredRulesCollapse` (subtests `within one object` / `across two objects`), which asserts the retained count, that the emptied object is the *second* one, and that nothing is left unplaced.

**Separator characters and multibyte values.** Added `TestKeyResistsSeparatorsAndMultibyteValues`, which sweeps values containing every encoding character (`= : ; [ ] ,`), values crafted to look like the encoding itself (`cidr=1:a;`, `=1:a;`, `1:a`), CJK, a precomposed-vs-decomposed pair (`é` as U+00E9 vs `e`+U+0301 — must stay distinct, since keys are byte-exact and deliberately not Unicode-normalized), and 1500-byte values including a same-prefix pair. Each is applied to `CIDR`, `Except`, and `DomainName`, asserting all keys are distinct and each is deterministic.

**nil vs empty `Except`.** Asserted equivalent, same as the existing `Ports` case.

**Format in godoc.** Added, and I generated it from the implementation rather than writing it by hand so it can't drift:

```
cidr=13:10.200.0.0/16;domain=0:;action=4:Deny;ports=1[proto=3:TCP;port=80;endport=nil;,];
```

## Answerable from the code

**Debug log when a key matched but content was replaced.** This rests on a false premise for the cluster path: `ClusterEndpointInfo` has exactly four fields — `CIDR`, `Ports`, `DomainName`, `Action` — and the key covers all four. With an injective encoding, a key match therefore *implies* content equality, so there is no "matched but replaced" event to log; the line would be unreachable. `DeepCopy` there is defense-in-depth against future field additions, not a live bug fix.

One honest caveat the review didn't raise: on the **namespaced** path, `getEndpointInfoKey` hashes `DomainName` *instead of* `CIDR`+`Except` in the FQDN branch (pre-existing behaviour I preserved). So a key match implies equality only under the documented CIDR/DomainName mutual exclusivity. That branch is the one place where the substitution could do real work, which is another reason to keep it.

**`slices.IsSorted(vals)` requires an ordered type.** `NetworkAddress` is `type NetworkAddress string`, so this is fine — and it's compile-time enforced, so it cannot regress silently.

**`comparePorts` / `cmp.Compare` / `SortFunc` toolchain support.** `go.mod` already declares `go 1.26.0`, well past the 1.21 floor these APIs need. No compatibility gap and nothing to update.

**`optInt32` encoding nil as `nil;`.** Consistent by construction: `nil;` cannot be confused with any decimal rendering, and nil vs empty slices are covered by tests on both `Ports` and `Except`.

**`slices.Grow` comment.** Skipped deliberately — the repo owner asked me to keep comments minimal, and the call is one line whose intent is evident. The sizing is a deliberate over-estimate to make the common case a single allocation; the measured effect was 16-port keys dropping from 9 allocs / 2336 B to 4 allocs / 1152 B.

---

Final state: `go vet ./...` and `go test ./...` clean, 12 tests plus a benchmark in `hashkey_test.go`, +472/−57.

haouc:
Addressed the FQDN gap I flagged above, and found the same bug in a second function.

## The problem

`getEndpointInfoKey` branched on `DomainName`, so in the FQDN case `CIDR` and `Except` were never hashed:

```go
if info.DomainName != "" {
    w.str("domain", string(info.DomainName))
} else {
    w.str("cidr", string(info.CIDR))
    w.strs("except", info.Except)
}
```

Two rules differing only in `CIDR` would therefore share an identity — the same class of bug this PR fixes, just reached by omission rather than by ambiguous encoding.

`getCombineKey` had it too, and inconsistently: it branched the same way on `DomainName`/`CIDR` but hashed `Except` in **both** branches. So the two key functions disagreed about whether `Except` is part of a rule's identity.

## The fix

Both functions now hash every field unconditionally, no branch:

```go
w.str("cidr", string(info.CIDR))
w.strs("except", info.Except)
w.str("domain", string(info.DomainName))
w.ports(info.Ports)   // getCombineKey omits Ports by design — it merges across them
```

Field tagging is what makes the branch unnecessary: `{CIDR: "example.com"}` and `{DomainName: "example.com"}` encode differently because the tags differ, so there is no need to select one field to disambiguate the other. This also brings the namespaced path in line with `getClusterEndpointInfoKey`, which already hashed all fields unconditionally.

`getCombineKey` additionally moves off unprefixed `|` concatenation onto the same length-prefixed writer, and its hand-rolled `Except` sort is gone (the writer canonicalizes). That removed the last use of `sort` in the file.

## Why this is behavior-preserving

I checked both producers of FQDN endpoints before changing anything — `applicationnetworkpolicy_resolver.go:178` and `clusternetworkpolicy_resolver.go:303`. Both construct FQDN endpoints with only `DomainName` and `Ports` (plus `Action` for the cluster type), leaving `CIDR` and `Except` at their zero values:

```go
endpoints = append(endpoints, policyinfo.EndpointInfo{
    DomainName: domain,
    Ports:      portList,
})
```

So the mutual-exclusivity invariant holds at every production site today, which means the change yields identical equivalence classes for every currently reachable input — and stops depending on an invariant that nothing enforces. `getCombineKey` is in-memory only and never persisted, so it carries no migration cost at all.

## Verification

New `TestFQDNKeyCoversCIDRAndExcept` asserts distinct keys for domain-only vs domain+CIDR, differing CIDRs, and domain-only vs domain+Except, against **both** key functions, plus that the domain and CIDR fields are not interchangeable.

Restoring the old branching fails it as expected:

```
--- FAIL: TestFQDNKeyCoversCIDRAndExcept
    domain only vs domain+cidr: share an endpoint key
    domain only vs domain+cidr: share a combine key
    differing cidr: share an endpoint key
    differing cidr: share a combine key
    domain only vs domain+except: share an endpoint key
```

Five of the six assertions fire. The sixth — `domain+except` on the combine key — does not, which is exactly the asymmetry described above: `getCombineKey` already hashed `Except` unconditionally, so only `getEndpointInfoKey` was dropping it.

`go vet ./...` and `go test ./...` clean, including `pkg/resolvers`, which exercises `combineRulesEndpoints`. Now +519/−82.

pkg/policyendpoints/clusterpolicyendpoints_chunking.go
```diff
@@ -102,11 +100,12 @@ func (m *policyEndpointsManager) processExistingClusterPolicyEndpoints(
 			existingCPEs[i].Spec.Egress = egEndpointList
 			existingCPEs[i].Spec.PodSelectorEndpoints = podSelectorEndpointList
 			potentialDeletes = append(potentialDeletes, existingCPEs[i])
-		} else if len(existingCPEs[i].Spec.Ingress) != len(ingEndpointList) ||
-			len(existingCPEs[i].Spec.Egress) != len(egEndpointList) ||
-			len(existingCPEs[i].Spec.PodSelectorEndpoints) != len(podSelectorEndpointList) ||
+		} else if !equality.Semantic.DeepEqual(existingCPEs[i].Spec.Ingress, ingEndpointList) ||
```
zac-nixon, line 103: Semantic deep equal feels like overkill here, specifically because we are already calculating the ruleKey above. Can you fold this into the upper loop?
```
ingChanged := false
ingEndpointList := make([]policyinfo.ClusterEndpointInfo, 0, len(existingCPEs[i].Spec.Ingress))
for _, ingRule := range existingCPEs[i].Spec.Ingress {
    ruleKey := m.getClusterEndpointInfoKey(ingRule)
    desired, exists := ingressEndpointsMap[ruleKey]
    if !exists {
        ingChanged = true // dropped, or a duplicate key already consumed
        continue
    }
    if !clusterEndpointInfoEqual(ingRule, desired) {
        ingChanged = true // same key, different representation (e.g. port order)
    }
    ingEndpointList = append(ingEndpointList, *desired.DeepCopy())
    delete(ingressEndpointsMap, ruleKey)
}
```

pkg/policyendpoints/hashkey.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func newHashKeyWriter() *hashKeyWriter {
+	return &hashKeyWriter{buf: make([]byte, 0, 128)}
+}
+
+func (w *hashKeyWriter) str(tag, val string) {
+	w.buf = append(w.buf, tag...)
+	w.buf = append(w.buf, '=')
+	w.buf = strconv.AppendInt(w.buf, int64(len(val)), 10)
```
zac-nixon, line 31: why do we need to include the length of the value in the hash key?

zac-nixon:
https://github.com/aws/amazon-network-policy-controller-k8s/pull/260

PR #263, opened 2026-09-16
fix: use delimiter in hash to denote different hashable objects

zac-nixon:
**What type of PR is this?**

bug

n/a

The current rule hashing algorithm doesn't use delimiters, which means it's possible that two fields can be concatenated to form what looks like an identical hash, but for objects that are actually different. This PR also uses itoa to convert an integer into a hashable object, rather than a []rune, which is lossy.

I added unit tests that validate the hashing behavior.

**Automation added to e2e**:

No this will not break downgrades as the hash is relative to the running process.

no


parlakisik:
@zac-nixon  did you able to validate this change in the cluster ?

Makefile
```diff
@@ -219,3 +219,4 @@ deploy-controller-on-dataplane: ## Deploys the Network Policy controller on an e
 deploy-and-test: ## Deploys the Network Policy controller on an existing cluster and runs cyclonus tests. Call with CLUSTER_NAME=<name of the cluster> and NP_CONTROLLER_IMAGE=<Image URI> 
 	$(MAKE) deploy-controller-on-dataplane NP_CONTROLLER_IMAGE=$(NP_CONTROLLER_IMAGE)
 	$(MAKE) run-cyclonus-test CLUSTER_NAME=$(CLUSTER_NAME)
+
```
parlakisik: nit: space

pkg/policyendpoints/manager_test.go
```diff
@@ -1262,6 +1262,59 @@ func Test_getEndpointInfoKey(t *testing.T) {
...
+	// Delimiter: adjacent-digit boundaries must not alias.
+	assert.NotEqual(t,
+		m.getEndpointInfoKey(rule(portRange(1, 23))),
+		m.getEndpointInfoKey(rule(portRange(12, 3))),
+		"range 1-23 must not hash the same as range 12-3")
+
+	// Delimiter: CIDR and except boundaries must be unambiguous.
+	assert.NotEqual(t,
+		m.getEndpointInfoKey(policyinfo.EndpointInfo{CIDR: "10.0.0.0/8", Except: []policyinfo.NetworkAddress{"10.1.0.0/16", "10.2.0.0/16"}}),
+		m.getEndpointInfoKey(policyinfo.EndpointInfo{CIDR: "10.0.0.0/810.1.0.0/16", Except: []policyinfo.NetworkAddress{"10.2.0.0/16"}}),
```
parlakisik, line 1309: is this cidr correct ?

zac-nixon: It's not a valid CIDR, but this test is just ensuring that two fields are correctly delimited (e.g. the CIDR and except CIDRs are not mashed together)

parlakisik: oh got it . validating the collusion

Makefile
zac-nixon: fixed

pkg/policyendpoints/clusterpolicyendpoints_chunking.go
```diff
@@ -277,22 +278,28 @@ func (m *policyEndpointsManager) getClusterEndpointInfoFromHashes(hashes []strin
...
+	hasher.Write([]byte(info.DomainName))
+	hasher.Write([]byte(delimiter))
+	hasher.Write([]byte(info.Action))
+	hasher.Write([]byte(delimiter))
 	for _, port := range info.Ports {
 		if port.Protocol != nil {
-			hasher.Write([]byte(string(*port.Protocol)))
+			hasher.Write([]byte((*port.Protocol)))
 		}
+		hasher.Write([]byte(delimiter))
```
parlakisik, line 294: if we add delimiter . the 80 ,443 and 443 ,80 will create different hashes. can this affect any mechanism

zac-nixon: this is existing behavior, [80, 443] and [443, 80] already created different hashes. We can also fix that, but this fixed was targeted at fixing hashes that incorrectly collide.

parlakisik:
lgtm

parlakisik (approved):
some issues related with previous issues

zac-nixon:
https://github.com/aws/amazon-network-policy-controller-k8s/pull/260

